Issue #40, opened 2015-03-12
PsByFlag is not tested


yegor256:
There is no unit test for `PsByFlag` class, let's create one

davvd:
@yegor256 right, I added the `bug` tag

davvd:
@yegor256 thanks for the report, I topped your acc for 15 mins, payment ID `000-02320972`

davvd:
@ikhvostenkov it's yours now, please proceed keeping in mind [our principles](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Feel free to ask any technical questions right here in the ticket... The budget of this issue is **30 mins**, which is exactly how much will be paid when the task is *done* (see [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for explanation)

ikhvostenkov:
@davvd I've done with the test. I have 3 questions:
1. Do you have any coding rules convention doc? I was trying to follow coding rules in other tests, but could be not perfect.
2. JUnit is absent in pom.xml. Tests will not run without this dependency. I do not understand how it works for instance on building machine. I've just added dependency locally in my IDE to run it.
 3. Shall I open pull request with my changes?

ikhvostenkov:
One more point: actually test was already created and covered ~60%, I've modified it to reach 100% coverage.

yegor256:
@ikhvostenkov just run `mvn clean install -Pqulice` on command line and you will see our coding conventions (we use qulice.com for that). JUnit is coming from parent pom, see http://www.yegor256.com/2015/02/05/jcabi-parent-maven-pom.html

Yes, just submit a pull request when you're ready, see http://www.yegor256.com/2014/04/15/github-guidelines.html

davvd:
@ikhvostenkov the task is your hands for the last 15 days.. keep in mind that if it's not closed in the next 48 hours, it will be re-assigned to someone else, see [No Obligations](http://www.xdsd.org/2014/04/13/no-obligations-principle.html) principle. [This article](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html) should help if you're stuck. -30 added to your rating, at the moment it is: [+0](http://www.netbout.com/b/35232?open=rating)

ikhvostenkov:
@yegor256 Could you please close it if you do not have any comments, due to pull request has been successfully merged?

yegor256:
@ikhvostenkov many thanks!

davvd:
[...]

Issue #76, opened 2015-03-25
broken tests PsCookieTest, TsAuthTest, RsFlashTest 


dmzaytsev:
build project fails when system's default locale is ru_RU.

"java.lang.IllegalArgumentException: header line of HTTP response "Set-Cookie: foo=urn%3Atest%3A99;Path=/;Expires=Пт, 27 мар 2015 00:34:57 GMT" doesn't match "[a-zA-Z\-]+:\p{Print}+" regular expression, but it should, according to RFC 7230"

Caused by using  locale specific's date format in PsCookie.exit

yegor256:
@davvd this is definitely a bug

davvd:
> @davvd this is definitely a bug

@yegor256 got it, tag `bug` is here

davvd:
@d-molotchko it's your task now, the budget is **30 mins**, go ahead

davvd:
@dmzaytsev thanks for the ticket, your account was topped for 15 mins, payment 54211899

dmzaytsev:
@d-molotchko thanks!

krzyk:
@yegor256 could you release new version of takes with this fix?

yegor256:
@rultor release, tag is `0.11.2`

davvd:
@d-molotchko I'm not sure who actually completed this task, since there were no messages from you above (correct me if I'm wrong)

d-molotchko:
@davvd I completed this task

yegor256:
@d-molotchko in the future, always post a short message into the ticket, explaining where you fixed the problem (in what pull request) and what exactly was done

yegor256:
@davvd please add extra 30 mins to @d-molotchko

d-molotchko:
@yegor256 ok

davvd:
> @davvd please add extra 30 mins to @d-molotchko 

@yegor256 I've paid 30 mins to @d-molotchko in transaction `54668412`

PR #81, opened 2015-03-26
 PsByFlag is not tested #40

ikhvostenkov:


davvd:
Many thanks, I will find someone to review it before we merge

davvd:
@longtimeago review this one, please

longtimeago:
@ikhvostenkov Thank you for the PR
Have you tried to run maven build with qulice profile enabled?
```
mvn clean install -Pqulice
```
We don't accept code with build errors

ikhvostenkov:
@longtimeago I am not able to build project due to https://github.com/yegor256/takes/issues/76 I have only Russian and German locale on my side. With both PsCookieTest, RsFlashTest and TsAuthTest are failed.

longtimeago:
@ikhvostenkov well ... it means that we have to postpone this pull request until #76 is done.

ikhvostenkov:
@longtimeago my bad, actually I can skip unit test and run only qulice verification. I've done it and will fix violations.

longtimeago:
@ikhvostenkov or, you can do following:
1) Ignore failing tests
2) Build the project to be sure that no checkstyle (etc) errors happen
3) Enable failing tests back and push the changes

longtimeago:
@ikhvostenkov you had before :)

ikhvostenkov:
@longtimeago yeah=)

ikhvostenkov:
@longtimeago redone according with qulice rules

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -23,34 +23,149 @@
...
     }
 
+    /**
+     * Checks PsByFlag equality.
+     *
+     * @throws Exception If some problem inside
+     */
+    @Test
+    public void equalsAndHashCodeEqualTest() throws Exception {
+        MatcherAssert.assertThat(
```
longtimeago: consider using ```nl.jqno.equalsverifier``` to test ```equals```

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -23,34 +23,149 @@
...
+        );
+    }
+
+    /**
+     * Checks PsByFlag inequality.
+     *
+     * @throws Exception If some problem inside
+     */
+    @Test
+    public void equalsAndHashCodeNotEqualTest() throws Exception {
```
longtimeago: same here

longtimeago:
@ikhvostenkov See several remarks above

ikhvostenkov:
@longtimeago With nl.jqno.equalsverifier test will not pass because of following error: "Transient field flag should not be included in equals/hashCode contract". I actually do not understand why flag and passes fields are transient, class PsByFlag is not serializable at all as I can see. Can I remove transient from these fields or is there any sense to left them transient?

longtimeago:
@ikhvostenkov I had long discussion with Yegor about why fields should be transient. From his explanation: the main idea is that even if somebody decided to serialize the class these fields would not be serialized. So, we can't remove ```transient``` from fields.
About equalsverifier, you can add ```suppress(Warning.TRANSIENT_FIELDS)``` to suppress some warnings.

ikhvostenkov:
@longtimeago ok, works now. One more question shall I put dependency to ```nl.jqno.equalsverifier``` to the root pom or shall it go to some dependent pom?

longtimeago:
@ikhvostenkov the only way to add a dependency in your case is to put it to project pom (with scope ```test``` !)

ikhvostenkov:
@longtimeago thank you. All changes have been added.

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -23,34 +23,125 @@
...
+ *
  * @author Yegor Bugayenko (yegor@teamed.io)
  * @version $Id$
  * @since 0.10
  */
 public final class PsByFlagTest {
+    /**
+     * Testable PsByFlag object.
+     */
+    private static PsByFlag psbyflag;
```
longtimeago: why static?

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -23,34 +23,125 @@
...
+    /**
+     * Key.
+     */
+    private static final String KEY = "some-key";
```
longtimeago: can't we just use different key in each test?

longtimeago:
@ikhvostenkov sorry, final remarks from me

src/test/java/org/takes/facets/auth/PsByFlagTest.java
ikhvostenkov: @longtimeago I've used static here due to ```transient``` fields ```flag``` and ```passes``` and you are right this is not necessary here, I will change it to ```transient```.

src/test/java/org/takes/facets/auth/PsByFlagTest.java
ikhvostenkov: @longtimeago sure, we can, I will redo this as well.

ikhvostenkov:
@longtimeago corrections have been added

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -23,34 +23,122 @@
...
+    /**
+     * HTTP request method.
+     */
+    private static final String METHOD = "GET";
+
+    /**
+     * Mocked Identity object used to test exit method.
+     */
+    @Mock
+    private static Identity identity;
```
longtimeago: looks like it's not the end ...
this can be in-lined to the test
Basically, we avoid using static fields and methods.

ikhvostenkov:
@longtimeago agree with ```METHOD``` - will use ```POST``` and ```GET``` to satisfy qulice rules. With ```Identity``` I do not agree, from my perspective mocked objects should be defined as fields, due to you can reuse it in your tests, but agree ```static``` modifier could be removed. If you are using mocked objects inside each test or prefer real objects, I will redo it as well, just say what is a good practice in our project in terms ob mocked objects.

longtimeago:
@ikhvostenkov we prefer to keep mocks inside the tests. If exactly the same mocks used in many tests we create a method to produce a mock and call it from the test. Or mock injection (as you proposed).
The other option - use fake instead of mock, but it's not the case.
So, in this particular case, ```Identity``` is used only ones and thus it should be inlined.

ikhvostenkov:
@longtimeago ok, got it and redid accordingly.

longtimeago:
looks like github still suffering from DDoS

longtimeago:
@rultor please merge

yegor256:
@ikhvostenkov I don't understand why do we need to share this `psbyflag` variable across all unit tests? why can't we create a new instance in each test method? test methods should be as isolated as possible and as independent from each other as possible. Besides that, the name `psbyflag` is absolutely wrong. The variable should **not** know anything about the actual implementation type. Call it `pass` and use type `Pass` for it.

ikhvostenkov:
@yegor256 I shared ```psbyflag``` to do not duplicate code in each method. Test methods are independent and isolated in any case, there is no object state sharing among test methods. ```@Before``` is called before each test method call in test. I will redo with variable type.

ikhvostenkov:
@yegor256 Does my explanation work for you? Can we accept this pull request?

yegor256:
@ikhvostenkov they are independent for compiler, but they are connected to each other for the reader of the code. When I read your test class I understand that there is something similar between these methods, since they share the same variable. It's an anti-pattern in general (even though it's promoted by JUnit), see http://stackoverflow.com/questions/333682/unit-testing-anti-patterns-catalogue

ikhvostenkov:
@yegor256 There is something similar between these methods - they all working(testing) with object under test referenced by same variable. I was not able to find current approach as an anti-pattern using link above.

yegor256:
@ikhvostenkov but why this variable is the same? you're testing different functionality in different methods, why object under test is the same?

ikhvostenkov:
@yegor256 Java object is not the same, java class (object under test) is the same represented with the same variable. Why we need to create n-times object by hands coping and pasting the same code in n test methods instead of using advantages of JUnit and ```Before``` annotation which will do all "dirty" work for us? + Lets imagine in all n test methods, class with non-default constructor is used, and we decided to change this constructor -> you have to change it in each n test methods, but if you are using ```Before```, you will change it in one place.

yegor256:
@ikhvostenkov that's exactly the point. We don't want to have an ability to change one place and affect all test methods. We want each method to act as a standalone mini-program that manipulates with its own objects. If we need to share some functionality between them, we should use private static methods. `@Before` and `@BeforeClass` JUnit annotations are really a bad idea.

ikhvostenkov:
@yegor256 okay, redone.

yegor256:
@rultor merge pls

longtimeago:
@rultor merge please

longtimeago:
@ikhvostenkov please, merge from master

longtimeago:
@ikhvostenkov ping

ikhvostenkov:
@longtimeago sorry, I was on vacations, merged now.

longtimeago:
@rultor merge pls

longtimeago:
@ikhvostenkov no problem, thanks!

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -46,11 +53,64 @@ public void skipsIfNothingFound() throws IOException {
...
+        final Response response = new RsWithStatus(
+            new RsWithType(
+                new RsWithBody("<html>This is test response</html>"),
+                    "text/html"
+            ),
+            200
+        );
+        MatcherAssert.assertThat(
+            new PsByFlag(
+                ImmutableMap.of(
```
yegor256: you can use `PsByFlag.Pair` instead

yegor256:
@ikhvostenkov a small comment from me above.. also, the branch is out of sync with master

src/test/java/org/takes/facets/auth/PsByFlagTest.java
ikhvostenkov: @yegor256 I've used here ```ImmutableMap``` to cover with tests creating ```PsByFlag``` objects using constructor with map entries.

src/test/java/org/takes/facets/auth/PsByFlagTest.java
```diff
@@ -46,11 +53,64 @@ public void skipsIfNothingFound() throws IOException {
...
+                    "key", (Pass) new PsFake(true)
```
yegor256: @ikhvostenkov what this class-casting is for? we're not using class casting in the framework

src/test/java/org/takes/facets/auth/PsByFlagTest.java
yegor256: @ikhvostenkov got it, right

ikhvostenkov:
@yegor256 Merged with upstream, should be fine now.

src/test/java/org/takes/facets/auth/PsByFlagTest.java
ikhvostenkov: @yegor256 Because maps are with different types ```ImmutableMap.of("some-key", new PsFake(true));``` - this is ```Map<String, PsFake>``` map, but constructor is using ```Map<String, Pass>``` map. I know that custing  is not a good practice, but here I've used it to workaround limit of the API. I create object in the same method where I cast and we can be 100% sure that this cast will work and also casting to the parent type works implicitly. Also we can change constructor here to:
```Java
public PsByFlag(final Map<String, ? extends Pass> map) {
        this(PsByFlag.class.getSimpleName(), map);
    }
```

yegor256: @ikhvostenkov Yes, changing the ctor would be a better option than casting

ikhvostenkov: @yegor256 shall we open change request for constructor change? I think this is not good idea to do it inside this task.

ikhvostenkov: But actually change is minor and maybe we can do it here.

longtimeago: @ikhvostenkov @yegor256 Actually, you can use ImmutableMap without changing constructor and without any casting
```
new PsByFlag(ImmutableMap.<String, Pass>of("key", new PsFake(true)))
```

ikhvostenkov: @yegor256 shall I use solution from @longtimeago or change constructor?

yegor256: @ikhvostenkov I like what @longtimeago proposed

ikhvostenkov:
@yegor256 @longtimeago casting has been removed.

yegor256:
@rultor merge

longtimeago:
@rultor please  merge again

yegor256:
@rultor merge

davvd:
@elenavolokhova please, review this ticket for compliance with our [QA rules](http://www.teamed.io/qa.html)

davvd:
@rultor deploy now pls

elenavolokhova:
@longtimeago 
>davvd commented 23 days ago
Many thanks, I will find someone to review it before we merge

>longtimeago commented 20 days ago
looks like github still suffering from DDoS

 > Messages in a ticket always start with a name of a user they are addressed to.
 
 As far as you are the performer of this ticket you are responsible for its quality. Please make sure that everyone follows this rule in your tickets in the future. Ask them to fix issues before closing the ticket.

 There is no description provided. Pull request must briefly explain what was done in the branch, so that we can use this information later, for traceability. 

Please ask to fix this issue and make sure that everyone follows this rule in your tickets in the future.

longtimeago:
@ikhvostenkov please, update PR description with the link to the source issue and some short description of what was done. Thanks!

ikhvostenkov:
@longtimeago done.

longtimeago:
@ikhvostenkov thanks!

elenavolokhova:
@longtimeago @ikhvostenkov Thank you!

elenavolokhova:
@davvd Ticket passed QA review. Quality is acceptable.

davvd:
> @davvd Ticket passed QA review. Quality is acceptable.

@elenavolokhova thank you, let's try to make it "good" next time

davvd:
@longtimeago just added 10 mins to @elenavolokhova (for QA), payment ID is `55932952`. Thank you, I have added **34 mins** to your account in payment/transaction "`55932992`", time consumed: 448 hours and 45 mins. review comments (c=19) added as a bonus. +34 added to your rating, current score is: [+4685](http://www.netbout.com/b/35024?open=rating)

Issue #28, opened 2015-03-08
SLF4J logging


yegor256:
Let's create a collection of wrappers in `org.takes.facets.slf4j` for our basic interface (`Takes`, `Take`, `Fork`, etc), which will log basic parameters of methods called.

We should use them like this:

```java
new TsLogged(
  new TsFork(
    // my config...
  )
);
```

That should be enough to see everything that's going on with this takes.

davvd:
@yegor256 got it, tag `bug` is here

davvd:
@yegor256 many thanks, **15 mins** added to your acc for reporting this bug, pmt ID `000-bac20427`

yegor256:
we need at lease these classes:

```
TsLogged
BkLogged
FtLogged
TkLogged
PsLogged
```

davvd:
@dmzaytsev it's yours now, please proceed keeping in mind [our principles](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Feel free to ask any technical questions right here in the ticket

The budget here is **30 mins**, which is exactly how much time will be paid for, when the task is [completed](http://www.xdsd.org/2014/04/15/definition-of-done.html)

dmzaytsev:
@yegor256 could I use @com.jcabi.aspects.Loggable ?
It will be looks like this

    public final class TsLogged implements Takes {
       private final transient Takes origin;
       public TsLogged(final Takes takes) {
           this.origin = takes;
       }

       @Override
       @Loggable
       public Take route(final Request request) throws IOException {
           return this.origin.route(request);
       }
   }

yegor256:
@dmzaytsev no, we can't add any third-party libraries here... we should work directly with SLF4J API

dmzaytsev:
@yegor256 
TsLogged for Takes
BkLogged for Back ?
FtLogged for Front ?
TkLogged for Take
PsLogged for Pass 

right ?

yegor256:
@dmzaytsev exactly

dmzaytsev:
@yegor256 could you explain please how to make right ticket for puzzle ?
e.g. in this task I need to implement a few Loggers, I will make only one and create a puzzle for others. How to make right `@todo` and create a task (s) for other classes?

yegor256:
@dmzaytsev you create what you think you can create in the available time frame and all the rest mark leave un-implemented and add puzzles there like this:

```java
/**
[...]

PR #89, opened 2015-03-28
#28 SLF4J logging

dmzaytsev:
task #28 
created slf4j logger wrapper
created logger decorator TkLogged for Take including the unit test


davvd:
Thanks, let me find someone who can review this pull request

davvd:
@longtimeago this pull request is for you, please review

src/main/java/org/takes/facets/slf4j/LogWrap.java
```diff
@@ -0,0 +1,115 @@
...
+        this.level = lvl;
+        this.logger = LoggerFactory.getLogger(clazz);
+    }
+
+    /**
+     * Log message.
+     * @param format Format string
+     * @param param Parameters
+     */
+    // @checkstyle JavadocLocationCheck (1 line)
```
longtimeago: you can put suppression to the javadoc itself
```
/**
* Log message.
* @param format Format string
* @param param Parameters
* @checkstyle CyclomaticComplexityCheck (1 line)
*/
```
and remove JavadocLocationCheck suppression

src/main/java/org/takes/facets/slf4j/LogWrap.java
```diff
@@ -0,0 +1,115 @@
...
+    // @checkstyle CyclomaticComplexityCheck (1 line)
+    public final void log(final String format, final Object... param) {
+        if (this.level == LogWrap.Level.TRACE
```
longtimeago: consider using switch by ```LogWrap.Level``` with default block

src/main/java/org/takes/facets/slf4j/FtLogged.java
```diff
@@ -0,0 +1,73 @@
...
+     * Original front.
+     */
+    private final transient Front origin;
+
+    /**
+     * Ctor.
+     * @param front Original
+     */
+    public FtLogged(final Front front) {
+        this(front, LogWrap.Level.TRACE);
```
longtimeago: are you sure we need TRACE level by default?

src/test/java/org/takes/facets/slf4j/BkLoggedTest.java
```diff
@@ -0,0 +1,76 @@
...
+        ).accept(new Socket());
+        Mockito.verify(appender)
+            .doAppend(
+                Matchers.argThat(
+                    new ArgumentMatcher<LoggingEvent>() {
+                        @Override
+                        public boolean matches(final Object argument) {
+                            return ((LoggingEvent) argument)
+                                .getRenderedMessage()
+                                .matches("^.*accept.*$");
```
longtimeago: I think ```contains("accept") would be enough

src/test/java/org/takes/facets/slf4j/FtLoggedTest.java
```diff
@@ -0,0 +1,84 @@
...
+        );
+        Mockito.verify(appender)
+            .doAppend(
+                Matchers.argThat(
+                    new ArgumentMatcher<LoggingEvent>() {
+                        @Override
+                        public boolean matches(final Object argument) {
+                            return ((LoggingEvent) argument)
+                                .getRenderedMessage()
+                                .matches("^.*start.*$");
```
longtimeago: same here

longtimeago:
@dmzaytsev Thanks for the PR.. Basically it's good but it's TOO big. It's really hard to review such complex PRs. I don't think you've spent 30 min on it. Next time, please, implement something simple to start and create todos in the code.
Also, see several remarks in the code above

src/main/java/org/takes/facets/slf4j/LogWrap.java
dmzaytsev: I think switch with if in each block will be less readable. isn't it?

src/main/java/org/takes/facets/slf4j/FtLogged.java
dmzaytsev: @longtimeago I think logs every method's call with parameters and return values will be too big. It could be used for debug only. Level.DEBUIG will be better ?

src/test/java/org/takes/facets/slf4j/BkLoggedTest.java
dmzaytsev: @longtimeago Could you suggest please any TypeSafeMatcher for this case ?
I don't know more suitable matcher to check mock arguments.

dmzaytsev:
@longtimeago Thanks for review, you are right I should been split the task.
I asked some questions above

src/main/java/org/takes/facets/slf4j/FtLogged.java
longtimeago: I would store default level inside the ```LogWrap``` and make it ```DEBUIG``` instead of ```TRACE```

src/test/java/org/takes/facets/slf4j/BkLoggedTest.java
longtimeago: @dmzaytsev Well, I would not use the mock at all.
We can create our own implementation of appender (extend ```org.apache.log4j.AppenderSkeleton```) witch will store log entries for us and provide it to us by request.
Tis appender should be enclosed inside some other class, say, ```TestLog``` with can extend ```org.junit.rules.ExternalResource``` and override ```before()``` and ```after()``` methods, where appended can be initialized and closed.
In the end, usage will be very simple:
1) Declare new field in test class:
```
    @Rule
    public final transient TestLog testlog = new TestLog();
```
2) And used in some test:
        MatcherAssert.assertThat(
            this.testlog.log(),
            Matchers.containsString("Some string we are waiting in logs")
        );
```
And that's it.

But, I understand, you did a lot of work in one PR and I proposed you to create a puzzle for another developer to do these changes.
And again, please, avoid such huge PR in future. They are really hard to manage.

File: src/main/java/org/takes/facets/slf4j/LogWrap.java [hunk 6]
```diff
@@ -0,0 +1,114 @@
...
+        this.logger = LoggerFactory.getLogger(clazz);
+    }
+
+    /**
+     * Log message.
+     * @param format Format string
+     * @param param Parameters
+     * @checkstyle CyclomaticComplexityCheck (2 line)
+     */
+    public final void log(final String format, final Object... param) {
```
longtimeago: first of all, I'm not sure we need to check if some level is enabled for all levels. In general, check is meaningful for DEBUG and TRACE as almost all application have default log level INFO. Take a look at how it could look like:
```java
    public final void log(final String format, final Object... param) {
        switch (this.level) {
            case TRACE :
                if (this.logger.isTraceEnabled()) {
                    this.logger.trace(String.format(format, param));
                }
                break;
            case DEBUG:
                if (this.logger.isDebugEnabled()) {
                    this.logger.debug(String.format(format, param));
                }
                break;
            case INFO:
                this.logger.info(String.format(format, param));
                break;
            case WARN:
                this.logger.warn(String.format(format, param));
                break;
            case SEVERE:
                this.logger.error(String.format(format, param));
                break;
            default:
                throw new IllegalArgumentException("Unknown log level");

        }
    }
```
what do you think?

dmzaytsev: I think we need to check for all levels.
for example application sets INFO level, but in some condition it will generate huge logs. We could  disable logs via log4.properties, but String.format() for a lot objects  may be too expensive. We need check level before call log.info() to avoid cpu's overhead.

longtimeago: well ... there is more clever solution - delegate formatting to the underlying logger:
```java
        switch (this.level) {
            case TRACE :
                this.logger.trace(format, param);
                break;
            case DEBUG:
                this.logger.debug(format, param);
                break;
            case INFO:
                this.logger.info(format, param);
                break;
            case WARN:
                this.logger.warn(format, param);
                break;
            case SEVERE:
                this.logger.error(format, param);
                break;
            default:
                throw new IllegalArgumentException("Unknown log level");
        }
```

longtimeago: I also propose to use ERROR instead of SEVERE to be consistent with SLF4J

dmzaytsev:
@longtimeago review again please

longtimeago:
@dmzaytsev thank you!

pom.xml
```diff
@@ -198,6 +198,12 @@
             <artifactId>javax.el-api</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.7</version>
+            <scope>compile</scope>
```
yegor256: we should add `<optional>true</optional>` here, in case a user doesn't want to use SLF4J logging, this dependency won't be in his project

src/main/java/org/takes/facets/slf4j/BkLogged.java
```diff
@@ -0,0 +1,75 @@
...
+import java.net.Socket;
+import lombok.EqualsAndHashCode;
+import org.takes.http.Back;
+
+/**
+ * Logs Back.accept() calls.
+ *
+ * <p>The class is immutable and thread-safe.
+ * @author Dmitry Zaytsev (dmitry.zaytsev@gmail.com)
+ * @version $Id$
```
yegor256: let's add `since 0.11`

src/test/java/org/takes/facets/slf4j/BkLoggedTest.java
```diff
@@ -0,0 +1,80 @@
...
+    public void logsMessage() throws IOException {
+        final Appender appender = Mockito.mock(Appender.class);
+        Mockito.when(appender.getName()).thenReturn("MOCK");
+        Logger.getRootLogger().addAppender(appender);
+        new BkLogged(
+            new Back() {
+                @Override
+                @SuppressWarnings("PMD.UncommentedEmptyMethod")
+                public void accept(final Socket socket) throws IOException {
+                }
```
yegor256: just add `assert socket != null` to the method and remove PMD suppression

yegor256: or even better, do `socket.getOutputStream().close()`

src/test/java/org/takes/facets/slf4j/BkLoggedTest.java
```diff
@@ -0,0 +1,80 @@
...
+            }, LogWrap.Level.WARN
+        ).accept(new Socket());
+        Mockito.verify(appender)
+            .doAppend(
```
yegor256: you can put `.doAppend(` to the previous line, will save you one indentation level

src/test/java/org/takes/facets/slf4j/FtLoggedTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void logsMessage() throws IOException {
+        final Appender appender = Mockito.mock(Appender.class);
+        Mockito.when(appender.getName()).thenReturn("MOCK");
+        Logger.getRootLogger().addAppender(appender);
+        new FtLogged(
+            new Front() {
+                @Override
+                @SuppressWarnings("PMD.UncommentedEmptyMethod")
+                public void start(final Exit exit) throws IOException {
+                }
```
yegor256: same here, add `assert exit != null`

yegor256:
@dmzaytsev the first question is why did you decided to implement all of them in one go? You could have done just one and create puzzles for all others. You obviously spent too much time on this ticket and created too much code, which is difficult to review and modify.

Second, I don't like the idea behind `LogWrap`. In general, using inheritance in order to share some functionality between objects is an anti-pattern. Instead, use encapsulation. Make an instance of `LogWrap` (I would rename it to `Target.Slf4j`) and encapsulate it in every class. This will also make testing way easier, since you will be able to mock this `Target` interface and intercept all logging calls.

Third, try to run `pdd -f /dev/null`, you will see issues with your PDD puzzles.

dmzaytsev:
@longtimeago @yegor256 Let's try again

I'll create stubs for BkLogged,FtLogged,PsLogged likeTsLogged.

dmzaytsev:
@yegor256 I fixed your comments. Review the changes please

src/main/java/org/takes/facets/slf4j/Slf4j.java
```diff
@@ -0,0 +1,145 @@
...
+ * Slf4j target.
+ *
+ * <p>The class is immutable and thread-safe.
+ * @author Dmitry Zaytsev (dmitry.zaytsev@gmail.com)
+ * @version $Id$
+ * @since 0.11.2
+ */
+@EqualsAndHashCode(of = "logger")
+@SuppressWarnings("PMD.LoggerIsNotStaticFinal")
+public final class Slf4j implements Target {
```
yegor256: this class should not be public. nobody should see it outside of the package.

src/main/java/org/takes/facets/slf4j/Slf4j.java
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * Default log level.
+     */
+    public static final Slf4j.Level DEFAULT_LEVEL = Slf4j.Level.DEBUG;
+
+    /**
+     * Log levels.
+     */
+    public enum Level {
```
yegor256: why do we need this enum if SLF4J already has `Slf4j.Level`?

src/main/java/org/takes/facets/slf4j/Target.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Log target.
+ *
+ * <p>All implementations of this interface must be immutable and thread-safe.
+ * @author Dmitry Zaytsev (dmitry.zaytsev@gmail.com)
+ * @version $Id$
+ * @since 0.11
+ */
+public interface Target {
```
yegor256: no need to make it public, this interface only for this package

src/main/java/org/takes/facets/slf4j/TkLogged.java
```diff
@@ -0,0 +1,82 @@
...
+     * Log target.
+     */
+    private final transient Target target;
+
+    /**
+     * Ctor.
+     * @param take Original
+     * @param trget Log target
+     */
+    public TkLogged(final Take take, final Target trget) {
```
yegor256: I think we should make this ctor package-visible only and introduce another one, which will be visible to everybody, without `Target` argument

src/main/java/org/takes/facets/slf4j/TsLogged.java
```diff
@@ -0,0 +1,52 @@
...
+ *  Please implement it like TkLogged and don't forget about unit tests.
+ *  In addition, need to be implement classes BkLogged, FtLogged, PsLogged
+ *  which will make loggable interfaces Back, Front, Pass. They should all be
+ *  implemented also like TkLogged.
+ * @since 0.11.2
+ */
+public final class TsLogged implements Takes {
+    @Override
+    public Take route(final Request request) throws IOException {
+        return new TkEmpty();
```
yegor256: it's better to throw `UnsupportedOperationException` when something is not implemented yet

src/main/java/org/takes/facets/slf4j/package-info.java
```diff
@@ -0,0 +1,45 @@
...
+ * This is how you implement it
+ * in your "take":
+ *
+ * <pre> new TsLogged(
+ * new TsFork(
+ * // my config...
+ * ),
+ * new Slf4j(TsFork.class)
+ * );
+ * </pre>
```
yegor256: what happened with indentations in this snippet? it looks rather unreadable

src/test/java/org/takes/facets/slf4j/TkLoggedTest.java
```diff
@@ -0,0 +1,65 @@
...
+        new TkLogged(
+            new TkText("test"),
+            new Target() {
+                @Override
+                public void log(final String format, final Object... param) {
+                    logs.add(
+                        MessageFormatter.arrayFormat(format, param).getMessage()
+                    );
+                }
+            }).act();
```
yegor256: see http://www.yegor256.com/2014/10/23/paired-brackets-notation.html

src/test/java/org/takes/facets/slf4j/TkLoggedTest.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * TkLogged can log message.
+     * @throws IOException If some problem inside
+     */
+    @Test
+    public void logsMessage() throws IOException {
+        final Collection<String> logs = new ArrayList<String>(1);
+        new TkLogged(
+            new TkText("test"),
+            new Target() {
```
yegor256: I would recommend to use `Mockito.verify` for this particular test, it will be shorter and easier to understand

src/test/java/org/takes/facets/slf4j/package-info.java
```diff
@@ -0,0 +1,32 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/**
+ * SLF4J wrappers , tests.
```
yegor256: why space before comma?

yegor256:
@dmzaytsev see some comments above

src/main/java/org/takes/facets/slf4j/Slf4j.java
dmzaytsev: @yegor256 SLF4J hasn't Level
Do you mean org.apache.log4j.Level ?

yegor256:
@dmzaytsev btw, I think `LogWrap` should be removed from the branch, right?

src/main/java/org/takes/facets/slf4j/Slf4j.java
yegor256: @dmzaytsev let's use `java.util.logging.Level`

dmzaytsev:
I fix all comments besides [this] (https://github.com/yegor256/takes/pull/89#discussion-diff-27631173)
I need your confirmation. If we use org.apache.log4j.Level, you'll have to add the log4j dependency in pom.xml. Is not it?

dmzaytsev:
@yegor256 sure, I alreday  removed LogWrap

dmzaytsev:
@yegor256 take a look the changes please

src/main/java/org/takes/facets/slf4j/Slf4j.java
```diff
@@ -0,0 +1,114 @@
...
+     * Logger.
+     */
+    private transient Logger logger;
+
+    /**
+     * Ctor.
+     * @param clazz Logger class
+     */
+    Slf4j(final Class<?> clazz) {
+        this(clazz, Slf4j.DEFAULT_LEVEL);
```
yegor256: it's better to do `this(clazz.getName())` here and get rid of this `DEFAULT_LEVEL` constant, being used only once

src/main/java/org/takes/facets/slf4j/TkLogged.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public TkLogged(final Take take) {
+        this(take, new Slf4j(TkLogged.class));
+    }
+
+    /**
+     * Print itself.
+     * @return Response
+     * @throws IOException If fails
+     * @todo #101 I expect implementations of Response and Take
```
yegor256: I think we have a bug/ticket already for this problem, right? Would be good to put a link to it, into this puzzle and explain exactly how you expect puzzle resolver to use this new feature we expect

src/main/java/org/takes/facets/slf4j/package-info.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Slf4j.
+ *
+ * <p>Classes in this package is a collection of wrappers for basic interfaces,
+ * which will log basic parameters of methods called.
+ * This is how you implement it
+ * in your "take":
+ *
+ * <pre> new TsLogged(
+ *           new TsFork(
```
yegor256: why eight spaces? :) see how similar snippets are documented in other classes, for example in `Takes`

yegor256:
@dmzaytsev we're almost there, a few more comments above

dmzaytsev:
@yegor256 review again please

src/main/java/org/takes/facets/slf4j/Slf4j.java
```diff
@@ -0,0 +1,114 @@
...
+ * @version $Id$
+ * @since 0.11.2
+ */
+@EqualsAndHashCode(of = "logger")
+@SuppressWarnings("PMD.LoggerIsNotStaticFinal")
+final class Slf4j implements Target {
+    /**
+     * Default log level.
+     */
+    public static final Level DEFAULT_LEVEL = Level.FINE;
```
yegor256: what is the point of this constant? why can't you use `Level.FINE` where it's needed?

src/test/java/org/takes/facets/slf4j/TkLoggedTest.java
```diff
@@ -0,0 +1,59 @@
...
+public final class TkLoggedTest {
+    /**
+     * TkLogged can log message.
+     * @throws IOException If some problem inside
+     */
+    @Test
+    public void logsMessage() throws IOException {
+        final Target target = Mockito.mock(Target.class);
+        new TkLogged(new TkText("test"), target).act();
+        Mockito.verify(
```
yegor256: I would format it like this:

```
Mockito.verify(target, Mockito.times(1)).log(
  // ...
);
```

yegor256:
@dmzaytsev two minor comments above

src/main/java/org/takes/facets/slf4j/Slf4j.java
dmzaytsev: reviewer suggested to store default log level inside wrapper
https://github.com/yegor256/takes/pull/89#discussion_r27353965

yegor256: @dmzaytsev yes, that's right, but you already store it inside wrapper, at `this.level`. This constant is redundant. Just use `Level.FINE` where you need it

src/main/java/org/takes/facets/slf4j/Slf4j.java
```diff
@@ -0,0 +1,114 @@
...
+    Slf4j(final Class<?> clazz) {
+        this(clazz.getName());
+    }
+
+    /**
+     * Ctor.
+     * @param name Logger name
+     */
+    Slf4j(final String name) {
+        this(name, Slf4j.DEFAULT_LEVEL);
```
yegor256: I mean, replace this line with `this(name, Level.FINE)`

dmzaytsev:
@yegor256 take a look

yegor256:
@rultor looks very good now, pls merge

dmzaytsev:
@yegor256 is  estimated time mandatory in puzzle description?

yegor256:
@dmzaytsev yep

dmzaytsev:
@yegor256 merge please

yegor256:
@rultor merge this

davvd:
@elenavolokhova please, let us know what do you think about this ticket, according to our [QA rules](http://www.teamed.io/qa.html)

davvd:
@rultor please deploy

elenavolokhova:
@davvd This ticket looks good from QA point of view.

davvd:
> @davvd This ticket looks good from QA point of view.

@elenavolokhova thanks, got it

davvd:
@longtimeago just added 10 mins to @elenavolokhova (for QA), payment ID is `54710489`; Much appreciated! **18 mins** was added to your account, payment ID `54710498`, time spent is 133 hours and 21 mins; you're getting extra minutes here (c=3); +18 added to your rating, at the moment it is: [+3757](http://www.netbout.com/b/35024?open=rating)

Issue #160, opened 2015-04-08
let's get rid of List.add()


yegor256:
We're using `add()` method in a few places, from `List` interface, mostly for concatenation of headers. Let's get rid of that and introduce a new class `org.takes.misc.Concat`, which implement `Iterable` and encapsulate two iterables inside:

```java
Iterable<String> list = new Concat<>(
  response.head(),
  Collections.singleton("Host: www.example.com")
);
```

davvd:
@yegor256 thanks, I added the "bug" tag

davvd:
@yegor256 thanks a lot for reporting, **15 mins** added to your acc, pmt ID `000-6b1af8f3`

davvd:
@super132 the task is for you now, follow [these guidelines](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Don't hesitate to ask any technical questions right here... The budget of this issue is **30 mins**, which is exactly how much will be paid when the task is *done* (see [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for explanation)

super132:
@yegor256 , may I know could you please point me an example that illustrating this bug? Do you mean to replace the List interface use in the whole code base or just replace the use of List interface specifically for concatenating headers only?

Thanks.

super132:
After digging into the codes more, my understanding is to replace the use of ```List``` as a tool to concatenate headers string by ```Concat``` class. If my understand is correct then I will start work on this.

Thanks.

yegor256:
@super132 this ticket is about concatenating headers only. You're right, we should use `Concat` class instead of manual concatenation of headers with `List.add()`. For example, this code in `RsWithHeaders`:

```
        final List<String> headers = new LinkedList<String>();
        for (final String hdr : head) {
            headers.add(hdr);
        }
        headers.add(header);
        return headers;
```

Should be replaced with:

```
return new Concat(head, Collections.singleton(headers));
```

super132:
PR #183 has been created for this task. Please review.

davvd:
@super132 check this ["no obligations principle"](http://www.xdsd.org/2014/04/13/no-obligations-principle.html).. This task is on your name for at least 15 days. If you can't close it within the next 48 hours I will have to assign someone else to it. [This article](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html) should help if you're stuck

[...]

PR #183, opened 2015-04-16
#160 let's get rid of List.add()

super132:
For task #160 , added ```Concat``` class and another utility class ```IterableTransform``` for header concatenation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,11 +136,6 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
         </dependency>
@@ -152,6 +147,11 @@
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -138,7 +138,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <scope>provided</scope>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.projectlombok</groupId>
@@ -152,11 +152,6 @@
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,11 @@
             <optional>true</optional>
         </dependency>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
         </dependency>
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -92,9 +92,7 @@ public final class Concat<T> implements Iterable<T> {
 
         @Override
         public boolean hasNext() {
-            final boolean first = this.left.hasNext();
-            final boolean second = this.right.hasNext();
-            return first || second;
+            return this.left.hasNext() || this.right.hasNext();
         }
 
         @Override
@@ -112,7 +110,9 @@ public final class Concat<T> implements Iterable<T> {
 
         @Override
         public void remove() {
-            throw new UnsupportedOperationException();
+            throw new UnsupportedOperationException(
+                "This iterable is immutable and cannot remove anything"
+            );
         }
     }
 }
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -30,9 +30,9 @@ import java.util.List;
 /**
  * Concat iterable.
  *
- * @author Yegor Bugayenko (yegor@teamed.io)
+ * @author Jason Wong (super132j@yahoo.com)
  * @version $Id$
- * @since 0.32.1
+ * @since 0.13.8
  */
 public final class Concat<T> implements Iterable<T> {
 
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -113,5 +113,10 @@ public final class Concat<T> implements Iterable<T> {
                 throw new NoSuchElementException();
             }
         }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
+        }
     }
 }
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -65,7 +65,7 @@ public final class Concat<T> implements Iterable<T> {
     }
 
     /**
-     * The concat iterator to traverse the input iterables as if they are 
+     * The concat iterator to traverse the input iterables as if they are
      * from one list.
      */
     private static class ConcatIterator<E> implements Iterator<E> {
@@ -81,9 +81,9 @@ public final class Concat<T> implements Iterable<T> {
         private final transient Iterator<E> right;
 
         /**
-         * Ctor. ConcatIterator traverses the element 
-         * @param aitr
-         * @param bitr
+         * Ctor. ConcatIterator traverses the element.
+         * @param aitr The first iterable to traverse
+         * @param bitr The second iterable to traverse
          */
         public ConcatIterator(final Iterator<E> aitr, final Iterator<E> bitr) {
             this.left = aitr;
@@ -92,15 +92,9 @@ public final class Concat<T> implements Iterable<T> {
 
         @Override
         public boolean hasNext() {
-            final boolean left = this.left.hasNext();
-            final boolean right = this.right.hasNext();
-            if (left) {
-                return true;
-            } else if (right) {
-                return true;
-            } else {
-                return false;
-            }
+            final boolean first = this.left.hasNext();
+            final boolean second = this.right.hasNext();
+            return first || second;
         }
 
         @Override
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -24,7 +24,6 @@
 package org.takes.misc;
 
 import java.util.Iterator;
-import java.util.NoSuchElementException;
 
 /**
  * Concat iterable.
@@ -100,10 +99,8 @@ public final class Concat<T> implements Iterable<T> {
             final E object;
             if (this.left.hasNext()) {
                 object = this.left.next();
-            } else if (this.right.hasNext()) {
-                object = this.right.next();
             } else {
-                throw new NoSuchElementException();
+                object = this.right.next();
             }
             return object;
         }
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -24,8 +24,7 @@
 package org.takes.misc;
 
 import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.NoSuchElementException;
 
 /**
  * Concat iterable.
@@ -37,9 +36,14 @@ import java.util.List;
 public final class Concat<T> implements Iterable<T> {
 
     /**
-     * Internal storage to hold the elements from iterables.
+     * Internal reference to hold the first elements from constructor.
      */
-    private final transient List<T> storage = new LinkedList<T>();
+    private final transient Iterable<T> left;
+
+    /**
+     * Internal reference to hold the second elements from constructor.
+     */
+    private final transient Iterable<T> right;
 
     /**
      * To produce an iterable collection combining a and b, with order of the
@@ -48,24 +52,66 @@ public final class Concat<T> implements Iterable<T> {
      * @param bitb Second iterable to conat
      */
     public Concat(final Iterable<T> aitb, final Iterable<T> bitb) {
-        this.concat(aitb);
-        this.concat(bitb);
+        this.left = aitb;
+        this.right = bitb;
     }
 
     @Override
     public Iterator<T> iterator() {
-        return this.storage.iterator();
+        return new ConcatIterator<T>(
+            this.left.iterator(),
+            this.right.iterator()
+        );
     }
 
     /**
-     * Adding an iterable into storage.
-     * @param itb Iterable to add
+     * The concat iterator to traverse the input iterables as if they are 
+     * from one list.
      */
-    private void concat(final Iterable<T> itb) {
-        final Iterator<T> itr = itb.iterator();
-        while (itr.hasNext()) {
-            this.storage.add(itr.next());
+    private static class ConcatIterator<E> implements Iterator<E> {
+
+        /**
+         * Internal reference for holding the first iterator form constructor.
+         */
+        private final transient Iterator<E> left;
+
+        /**
+         * Internal reference for holding the second iterator form constructor.
+         */
+        private final transient Iterator<E> right;
+
+        /**
+         * Ctor. ConcatIterator traverses the element 
+         * @param aitr
+         * @param bitr
+         */
+        public ConcatIterator(final Iterator<E> aitr, final Iterator<E> bitr) {
+            this.left = aitr;
+            this.right = bitr;
+        }
+
+        @Override
+        public boolean hasNext() {
+            final boolean left = this.left.hasNext();
+            final boolean right = this.right.hasNext();
+            if (left) {
+                return true;
+            } else if (right) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+
+        @Override
+        public E next() {
+            if (this.left.hasNext()) {
+                return this.left.next();
+            } else if (this.right.hasNext()) {
+                return this.right.next();
+            } else {
+                throw new NoSuchElementException();
+            }
         }
     }
-
 }
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -36,7 +36,7 @@ import java.util.List;
  */
 public final class Concat<T> implements Iterable<T> {
 
-	private final List<T> storage = new LinkedList<T>();
+	private final transient List<T> storage = new LinkedList<T>();
 	
 	public static interface Condition<T> {
 		/**
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -99,13 +99,15 @@ public final class Concat<T> implements Iterable<T> {
 
         @Override
         public E next() {
+            final E object;
             if (this.left.hasNext()) {
-                return this.left.next();
+                object = this.left.next();
             } else if (this.right.hasNext()) {
-                return this.right.next();
+                object = this.right.next();
             } else {
                 throw new NoSuchElementException();
             }
+            return object;
         }
 
         @Override
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -41,20 +41,10 @@ public final class Concat<T> implements Iterable<T> {
      */
     private final transient List<T> storage = new LinkedList<T>();
 
-    public interface Condition<T> {
-        /**
-         * Determine if an element should be added.
-         * 
-         * @param element The element in the iterables to examine.
-         * @return True to add the element, false to skip.
-         */
-        boolean add(T element);
-    }
-
     /**
      * To produce an iterable collection combining a and b, with order of the
      * elements in a first.
-     * 
+     *
      * @param aitb First iterable to concat
      * @param bitb Second iterable to conat
      */
@@ -69,25 +59,24 @@ public final class Concat<T> implements Iterable<T> {
      * 
      * @param aitb First iterable to concat
      * @param bitb Second iterable to conat
-     * @param cond
-     *            To determine which element in the iterables to add in the
-     *            final iterable.
+     * @param cond To determine which element to add in the final iterable
      */
-    public Concat(final Iterable<T> aitb, final Iterable<T> bitb, final Condition<T> cond) {
+    public Concat(final Iterable<T> aitb, final Iterable<T> bitb,
+            final Condition<T> cond) {
         this.concat(aitb, cond);
         this.concat(bitb, cond);
     }
 
     /**
      * Adding an iterable into storage with condition.
-     * 
+     *
      * @param itb Iterable to add
      * @param cond Condition to determine the element should be added
      */
     private void concat(final Iterable<T> itb, final Condition<T> cond) {
-        Iterator<T> itr = itb.iterator();
+        final Iterator<T> itr = itb.iterator();
         while (itr.hasNext()) {
-            T element = itr.next();
+            final T element = itr.next();
             if (cond.add(element)) {
                 this.storage.add(element);
             }
@@ -96,19 +85,29 @@ public final class Concat<T> implements Iterable<T> {
 
     /**
      * Adding an iterable into storage.
-     * 
+     *
      * @param itb Iterable to add
      */
     private void concat(final Iterable<T> itb) {
-        Iterator<T> itr = itb.iterator();
+        final Iterator<T> itr = itb.iterator();
         while (itr.hasNext()) {
             this.storage.add(itr.next());
         }
     }
 
     @Override
-    public Iterator<T> iterator() {
+    public final Iterator<T> iterator() {
         return this.storage.iterator();
+    }
+    
+    public interface Condition<T> {
+        /**
+         * Determine if an element should be added.
+         *
+         * @param element The element in the iterables to examine.
+         * @return True to add the element, false to skip.
+         */
+        boolean add(T element);
     }
 
 }
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -66,9 +66,9 @@ public final class Concat<T> implements Iterable<T> {
         this.concat(aitb, cond);
         this.concat(bitb, cond);
     }
-    
+
     @Override
-    public final Iterator<T> iterator() {
+    public Iterator<T> iterator() {
         return this.storage.iterator();
     }
 
@@ -99,7 +99,7 @@ public final class Concat<T> implements Iterable<T> {
             this.storage.add(itr.next());
         }
     }
-    
+
     public interface Condition<T> {
         /**
          * Determine if an element should be added.
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -44,7 +44,6 @@ public final class Concat<T> implements Iterable<T> {
     /**
      * To produce an iterable collection combining a and b, with order of the
      * elements in a first.
-     *
      * @param aitb First iterable to concat
      * @param bitb Second iterable to conat
      */
@@ -56,7 +55,6 @@ public final class Concat<T> implements Iterable<T> {
     /**
      * To produce an iterable collection, determined by condition, combining a
      * and b, with order of the elements in a first.
-     *
      * @param aitb First iterable to concat
      * @param bitb Second iterable to conat
      * @param cond To determine which element to add in the final iterable
@@ -74,7 +72,6 @@ public final class Concat<T> implements Iterable<T> {
 
     /**
      * Adding an iterable into storage with condition.
-     *
      * @param itb Iterable to add
      * @param cond Condition to determine the element should be added
      */
@@ -90,7 +87,6 @@ public final class Concat<T> implements Iterable<T> {
 
     /**
      * Adding an iterable into storage.
-     *
      * @param itb Iterable to add
      */
     private void concat(final Iterable<T> itb) {
--- a/src/main/java/org/takes/misc/Concat.java
+++ b/src/main/java/org/takes/misc/Concat.java
@@ -1,0 +1,93 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.takes.misc;
+
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * Concat iterable.
+ *
+ * @author Yegor Bugayenko (yegor@teamed.io)
+ * @version $Id$
+ * @since 0.32.1
+ */
+public final class Concat<T> implements Iterable<T> {
+
+	private final List<T> storage = new LinkedList<T>();
+	
+	public static interface Condition<T> {
+		/**
+		 * Determine if an element should be added.
+		 * @param element
+		 * @return
+		 */
+		boolean add(T element);
+	}
+	
+	/**
+	 * To produce an iterable collection combining a and b, with order of the elements in a first.
+	 * @param a
+	 * @param b
+	 */
+	public Concat(Iterable<T> a, Iterable<T> b) {
+		concat(a);
+		concat(b);
+	}
+	
+	/**
+	 * To produce an iterable collection, determined by condition, combining a and b, with order of the elements in a first.
+	 * @param a
+	 * @param b
+	 * @param cond
+	 */
+	public Concat(Iterable<T> a, Iterable<T> b, Condition<T> cond) {
+		concat(a, cond);
+		concat(b, cond);
+	}
+
+	private void concat(Iterable<T> a, Condition<T> cond) {
+		Iterator<T> i = a.iterator();
+		while(i.hasNext()) {
+			T element = i.next();
+			if(cond.add(element)) {
+				this.storage.add(element);
+			}
+		}
+	}
+
+	private void concat(Iterable<T> a) {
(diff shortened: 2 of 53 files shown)


super132:
@davvd , please review the pull request.

davvd:
@super132 Let me find a reviewer for this pull request, thanks for submitting it

davvd:
@krzyk review this one,please

krzyk:
@super132 please merge changes from master to this PR

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,113 @@
...
+ * @author Jason Wong (super132j@yahoo.com)
+ * @version $Id$
+ * @since 0.13.8
+ */
+public final class Concat<T> implements Iterable<T> {
+
+    /**
+     * Internal storage to hold the elements from iterables.
+     */
+    private final transient List<T> storage = new LinkedList<T>();
```
krzyk: @super132 why create a storage if we can reuse the original iterables for this? This will waste more memory.

src/main/java/org/takes/rs/RsWithHeaders.java
```diff
@@ -79,4 +78,13 @@ public InputStream body() throws IOException {
         );
     }
 
+    private static final class Action implements
```
krzyk: @super132 this should be an inner class of IterableTransformer

src/main/java/org/takes/misc/IterableTransform.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.List;
+
+/**
+ * Transform elements in an iterable into others.
+ *
+ * @author Jason Wong (super132j@yahoo.com)
+ * @version $Id$
+ * @since 0.13.8
+ */
+public class IterableTransform<T> implements Iterable<T> {
```
krzyk: @super132 maybe just `Transformer`

src/main/java/org/takes/misc/IterableTransform.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Internal storage.
+     */
+    private final transient List<T> storage = new LinkedList<T>();
```
krzyk: @super132 again, why create additional storage if we can iterate on the original one and transform in a lazy way

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param itb Iterable to add
+     */
+    private void concat(final Iterable<T> itb) {
+        final Iterator<T> itr = itb.iterator();
+        while (itr.hasNext()) {
+            this.storage.add(itr.next());
+        }
+    }
+
+    public interface Condition<T> {
```
krzyk: @super132 It would be better to move it into it's own file, and move `LcCondition` into it as inner class with `LowerCase`

src/main/java/org/takes/misc/LcCondition.java
```diff
@@ -0,0 +1,57 @@
...
+ * SOFTWARE.
+ */
+package org.takes.misc;
+
+import java.util.Locale;
+import org.takes.misc.Concat.Condition;
+
+/**
+ * Concat lower case string condition.
+ *
```
krzyk: @super132 please add information about prefix and how it is used

krzyk:
@super132 please see my comments above

src/main/java/org/takes/misc/Concat.java
super132: @krzyk , may I know if you mean to initialise it lazily or this storage is not even needed? If it is the later case then how could we reuse storage as this class should not holds any reference to the input iterables.

src/main/java/org/takes/misc/Concat.java
super132: @krzyk , thanks for the comment. I will change it in my next commit.

src/main/java/org/takes/misc/IterableTransform.java
super132: @krzyk , no problem. I will change it in my next commit.

src/main/java/org/takes/misc/IterableTransform.java
super132: @krzyk , do you mean to instantiate the ```List``` class when the supplied iterable is not empty? If it is the case then how are we going to deal with ```iterator()``` method if there is no element in the resultant iterable? Shouldn't ```Transformer``` and ```Concat``` always create new iterable object ?

src/main/java/org/takes/misc/LcCondition.java
super132: @krzyk , no problem. Will fix it in my next commit.

src/main/java/org/takes/rs/RsWithHeaders.java
super132: @krzyk , no problem. Will refactor this in my next commit.

super132:
@krzyk, thanks for your comments. I have updated some in the commit. I have some questions regarding to your comments and would you please help take a look of it? Thanks.

src/main/java/org/takes/misc/Concat.java
krzyk: @super132 why shouldn't this class hold references to the input iterables?

src/main/java/org/takes/misc/Transformer.java
```diff
@@ -0,0 +1,105 @@
...
+            this.storage.add(action.transform(itr.next()));
+        }
+    }
+
+    @Override
+    public final Iterator<K> iterator() {
+        return this.storage.iterator();
+    }
+
+    public interface Action<T, K> {
```
krzyk: @super132 I think here we should do as with `Condition`, so please move `Action` to a separate class and place the implementers as inner classes there.

src/main/java/org/takes/misc/Transformer.java
```diff
@@ -0,0 +1,105 @@
...
+         * @param element Element of the iterable
+         * @return Transformed element
+         */
+        K transform(T element);
+    }
+
+    /**
+     * Trimming action used with {@link Transformer}.
+     *
+     * @author Jason Wong (super132j@yahoo.com)
```
krzyk: @super132 we don't usually add tags (author, id, since) to inner classes

krzyk:
@super132 Please see my additional comments.
I'm also wondering if it wouldn't be good to name `Condition` and `Action`, just like it is named in Java 1.8 or Guava: `Predicate`(with test method) and `Function` (with apply method).

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Internal storage to hold the elements from iterables.
+     */
+    private final transient List<T> storage = new LinkedList<T>();
+
+    /**
+     * To produce an iterable collection combining a and b, with order of the
+     * elements in a first.
+     *
```
krzyk: @super132 generally we don't add new lines between method description and params (we do that only for class descriptions)

src/main/java/org/takes/misc/Condition.java
```diff
@@ -0,0 +1,78 @@
...
+     */
+    boolean add(T element);
+
+    /**
+     * Concat lower case string condition. This condition changes all
+     * characters to lower case and determine if the iterable should be
+     * concatenated by determine if the element contain the 'prefix'
+     * supplied.
+     *
+     * @author Jason Wong (super132j@yahoo.com)
```
krzyk: @super132 please remove tags from all inner classes

src/main/java/org/takes/misc/Transformer.java
super132: @krzyk , I missed this one. I will move it out as a separate class.

src/main/java/org/takes/misc/Concat.java
super132: @krzyk, I have 2 questions regarding to holding references:

1. How are we implement ```iterator()``` method? Do we need to write a brand new iterator which calls the iterators of the two iterables?
2. How are the use cases of ```Concat```? If there are more use cases that the input iterable will eventually be garbage collected then I think creating one new object would save more memory. I think when we use ```Concat```, we would be more interested in the end result rather than the input and the input would then be unreferenced. 
3. How would we handle ```remove()``` method that could be implemented by iterables? If the input iterable is modified by this method then it would affect the concatenated result if we rely on the reference to produce the output.

Thanks.

krzyk: @super132
For now let's leave it as it is and will see what @yegor256 thinks about current design.
 
1. You would create a new iterator that would iterate over the first iterable and after it finishes it would go to the other one
2. I don't know that.
3. Remove would just throw `UnsupportedOperationException` as it does in some iterators in JDK.

src/main/java/org/takes/rs/RsWithHeaders.java
```diff
@@ -62,15 +63,18 @@ public RsWithHeaders(final Response res,
...
+                @SuppressWarnings("unchecked")
+                public Iterable<String> head() throws IOException {
+                    return new Concat<String>(
+                        res.head(),
+                        new Transformer<String, String>(
+                            new Transformer<CharSequence, String>(
+                                (Iterable<CharSequence>) headers,
+                                new TransformAction.ToString()
+                            )
+                            , new TransformAction.Trim()
```
krzyk: @super132 coma, should be on the previous line

src/main/java/org/takes/rq/RqWithoutHeader.java
```diff
@@ -52,18 +53,16 @@ public RqWithoutHeader(final Request req, final CharSequence name) {
...
-                    for (final String header : req.head()) {
-                        if (!header.toLowerCase(Locale.ENGLISH)
-                            .startsWith(prefix)) {
-                            head.add(header);
-                        }
-                    }
-                    return head;
+                    return new Concat<String>(
+                        req.head(),
+                        Collections.EMPTY_LIST,
```
krzyk: @super132 I think it is better to use `Collections.emptyList()` method instead of static field

src/main/java/org/takes/misc/Transformer.java
```diff
@@ -0,0 +1,63 @@
...
+public class Transformer<T, K> implements Iterable<K> {
+
+    /**
+     * Internal storage.
+     */
+    private final transient List<K> storage = new LinkedList<K>();
+
+    /**
+     * Transform elements in the supplied iterable by the action supplied.
+     *
```
krzyk: @super132 please remove this new line

src/main/java/org/takes/misc/TransformAction.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param element Element of the iterable
+     * @return Transformed element
+     */
+    K transform(T element);
+
+    /**
+     * Trimming action used with {@link Transformer}.
+     */
+    class Trim implements
+        TransformAction<String, String> {
```
krzyk: @super132 this can be on the previous line

src/main/java/org/takes/misc/TransformAction.java
```diff
@@ -0,0 +1,66 @@
...
+ * Action for {@link Transformer} to perform actual transformation.
+ *
+ * @author Jason Wong (super132j@yahoo.com)
+ * @version $Id$
+ * @since 0.13.8
+ */
+public interface TransformAction<T, K> {
+    /**
+     * The transform action of the element of type T to K.
+     *
```
krzyk: @super132 please remove this new line

src/main/java/org/takes/misc/Condition.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @author Jason Wong (super132j@yahoo.com)
+ * @version $Id$
+ * @since 0.13.8
+ */
+public interface Condition<T> {
+
+    /**
+     * Determine if an element should be added.
+     *
```
krzyk: @super132 please remove this new line

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,103 @@
...
+            final T element = itr.next();
+            if (cond.add(element)) {
+                this.storage.add(element);
+            }
+        }
+    }
+
+    /**
+     * Adding an iterable into storage.
+     *
```
krzyk: @super132 please remove this new line

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return this.storage.iterator();
+    }
+
+    /**
+     * Adding an iterable into storage with condition.
+     *
```
krzyk: @super132 please remove this new line

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    public Concat(final Iterable<T> aitb, final Iterable<T> bitb) {
+        this.concat(aitb);
+        this.concat(bitb);
+    }
+
+    /**
+     * To produce an iterable collection, determined by condition, combining a
+     * and b, with order of the elements in a first.
+     *
```
krzyk: @super132 please remove this new line

src/main/java/org/takes/rs/RsWithoutHeader.java
```diff
@@ -52,18 +53,16 @@ public RsWithoutHeader(final Response res, final CharSequence name) {
...
-                    for (final String hdr : res.head()) {
-                        if (!hdr.toLowerCase(Locale.ENGLISH)
-                            .startsWith(prefix)) {
-                            head.add(hdr);
-                        }
-                    }
-                    return head;
+                    return new Concat<String>(
+                        res.head(),
+                        Collections.EMPTY_LIST,
```
krzyk: @super132 please use `Collections.emptyList()`

src/main/java/org/takes/rs/RsWithoutHeader.java
```diff
@@ -52,18 +53,16 @@ public RsWithoutHeader(final Response res, final CharSequence name) {
         super(
             new Response() {
                 @Override
+                @SuppressWarnings("unchecked")
```
krzyk: @super132 Is this required? I don't see any casting here.

src/main/java/org/takes/rs/RsWithStatus.java
```diff
@@ -112,17 +113,29 @@ public InputStream body() throws IOException {
...
-            }
-        }
-        return head;
+        return new Concat<String>(
+            Collections.singleton(
+                    String.format("HTTP/1.1 %d %s", status, reason)
+            ),
+            new Concat<String>(
+                    origin.head(),
+                    Collections.EMPTY_LIST,
```
krzyk: @super132 please use `Collections.emptyList()` here

src/main/java/org/takes/rs/RsWithStatus.java
```diff
@@ -100,6 +100,7 @@ public InputStream body() throws IOException {
      * @return Head
      * @throws IOException If fails
      */
+    @SuppressWarnings("unchecked")
```
krzyk: @super132 is this required?

krzyk:
@super132 please see my mostly styling comments, after those it looks like we can merge

src/main/java/org/takes/rs/RsWithoutHeader.java
super132: @krzyk, this is indeed needed as ```Collections.EMPTY_LIST``` is a generic List and ```Concat``` constructor expect an iterable with type ```String``` here.

src/main/java/org/takes/rs/RsWithoutHeader.java
super132: @krzyk, I tried using Collections.emptyList() but it actually returns List<Object>, which cannot cast back to Iterable<String> but raw type can do so.

src/main/java/org/takes/rs/RsWithHeaders.java
super132: @krzyk, no problem. Should this rule be added back to qulice?

src/main/java/org/takes/rq/RqWithoutHeader.java
super132: @krzyk, it cannot be used as it returns ```List<Object>``` type and cannot be casted back to ```Iterable<String>```

src/main/java/org/takes/misc/Transformer.java
super132: @krzyk, no problem.

src/main/java/org/takes/misc/Concat.java
super132: @krzyk, no problem. Thanks.

src/main/java/org/takes/misc/Concat.java
super132: @krzyk, no problem. It will be removed.

src/main/java/org/takes/rs/RsWithStatus.java
super132: @krzyk, yes. Same reason as above.

src/main/java/org/takes/rs/RsWithStatus.java
super132: @krzyk, same reason as above for not able to use the captioned method.

super132:
@krzyk, thanks for your comments. I have modified based on your advices and replied to your comments which I have difficulties in applying the changes. Thanks.

krzyk:
@super132 you can use `Collections.<String>emptyList()` and this way we can get rid of SuppressWarnings

super132:
@krzyk, thanks for your advice and I have modified the code to get rid of this.

krzyk:
@rultor merge pls

yegor256:
@super132 there are a few problems here. First, `Concat` is doing two things: concatenating and filtering. There should be two classes for that: `Concat` and `Select`. Second, `Transformer` is a bad name, let's name it `Transform` instead, see http://www.yegor256.com/2015/03/09/objects-end-with-er.html Once you fix that, we'll discuss the design of `Concat`. At the moment it's definitely wrong, but I hope it will become better once you take filtering out of it.

super132:
@yegor256 , Got it with single responsibility approach. I have revised the codes and would you please review it to see if it is better? Thanks.

src/main/java/org/takes/misc/Concat.java
yegor256: @super132 well, you're not solving the key problem with this implementation. At the moment `Concat` class is fetching items from all iterables during constructing. What if they have 1bln item each and the user will do: `new Concat(bln, bln).iterator().next();` You will fetch 2 bln items, while the user needs just one.

src/main/java/org/takes/misc/Condition.java
```diff
@@ -0,0 +1,72 @@
...
+ * @since 0.13.8
+ */
+public interface Condition<T> {
+
+    /**
+     * Determine if an element should be added.
+     * @param element The element in the iterables to examine.
+     * @return True to add the element, false to skip.
+     */
+    boolean add(T element);
```
yegor256: I would rename it to `fits()`

src/main/java/org/takes/misc/Condition.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Concat lower case string condition. This condition changes all
+     * characters to lower case and determine if the iterable should be
+     * concatenated by determine if the element contain the 'prefix'
+     * supplied.
+     */
+    class LowerCase implements Condition<String> {
```
yegor256: This class looks too specific for one particular case. I would not create it here, but instead make an anonymous class where it's needed

src/main/java/org/takes/misc/Select.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Adding an iterable into storage with condition.
+     * @param itb Iterable to add
+     * @param cond Condition to determine the element should be added
+     */
+    private void select(final Iterable<T> itb, final Condition<T> cond) {
+        final Iterator<T> itr = itb.iterator();
+        while (itr.hasNext()) {
```
yegor256: same problem here as in `Concat`. the idea of these classes is to fetch items from iterables **on-demand**, not on construction

src/main/java/org/takes/misc/Transform.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Transform elements in the supplied iterable by the action supplied.
+     * @param list Iterable to be transformed
+     * @param action The actual transformation implementation
+     */
+    public Transform(final Iterable<T> list,
+            final TransformAction<T, K> action) {
+        final Iterator<T> itr = list.iterator();
+        while (itr.hasNext()) {
```
yegor256: the same here. constructor should be fast and lightweight. in your case, it does all the heavy work of filtering, transforming, concatenating, etc.

yegor256:
@super132 a few more comments above

src/main/java/org/takes/misc/Condition.java
super132: @yegor256 , I tried using anonymous implementation in previous commits but qulice complained that the lines for the anonymous class is too long. Thus, I move it to here and there are 2 classes sharing the same logic.

src/main/java/org/takes/misc/Concat.java
super132: @yegor256, understood. I would change the implementation for it.

super132:
@yegor256 , I have revised the codes as per your comments. I left only one unchanged as limited by the number of lines of anonymous class by qulice. Thanks

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,118 @@
...
+        public ConcatIterator(final Iterator<E> aitr, final Iterator<E> bitr) {
+            this.left = aitr;
+            this.right = bitr;
+        }
+
+        @Override
+        public boolean hasNext() {
+            final boolean first = this.left.hasNext();
+            final boolean second = this.right.hasNext();
+            return first || second;
```
yegor256: this implementation is not effective, you will call both `hasNext()` methods always, even when the first call is sufficient to return true. should be `return hasNext() || hasNext()`

src/main/java/org/takes/misc/Concat.java
```diff
@@ -0,0 +1,118 @@
...
+                object = this.right.next();
+            } else {
+                throw new NoSuchElementException();
+            }
+            return object;
+        }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
```
yegor256: I think we should add some text here, to explain why we're rejecting `remove()` call. maybe something like `"this iterable is immutable, can't remove anything"`

src/main/java/org/takes/misc/Condition.java
```diff
@@ -0,0 +1,72 @@
...
+         * @param str The prefix to check
+         */
+        public LowerCase(final String str) {
+            this.prefix = str;
+        }
+
+        @Override
+        public boolean fits(final String element) {
+            return !element.toLowerCase(Locale.ENGLISH)
+                    .startsWith(this.prefix);
```
yegor256: why so many indentation spaces here?

src/main/java/org/takes/misc/Select.java
```diff
@@ -0,0 +1,159 @@
...
+
+        /**
+         * Ctor. ConcatIterator traverses the element.
+         * @param itr Iterator of the original iterable
+         * @param cond Condition to filter out elements
+         */
+        public SelectIterator(final Iterator<E> itr, final Condition<E> cond) {
+            this.condition = cond;
+            this.iterator = itr;
+            if (this.iterator.hasNext()) {
```
yegor256: again, you should not do this in ctor. what if the user never calls `next()`? you are still touching the iterable. it's wrong.

src/main/java/org/takes/misc/Select.java
```diff
@@ -0,0 +1,159 @@
...
+        @Override
+        public E next() {
+            if (this.hasNext()) {
+                final E result = this.current.get(HEAD);
+                this.current.remove(0);
+                if (this.iterator.hasNext()) {
+                    this.current.add(this.iterator.next());
+                }
+                return result;
+            } else {
```
yegor256: see http://www.yegor256.com/2015/01/21/if-then-throw-else.html

src/main/java/org/takes/misc/Transform.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    private final transient TransformAction<T, K> action;
+
+    /**
+     * Transform elements in the supplied iterable by the action supplied.
+     * @param itb Iterable to be transformed
+     * @param act The actual transformation implementation
+     */
+    public Transform(final Iterable<T> itb,
+            final TransformAction<T, K> act) {
```
yegor256: why 8 indentation spaces?

src/main/java/org/takes/misc/Transform.java
```diff
@@ -0,0 +1,112 @@
...
+        @Override
+        public boolean hasNext() {
+            return this.iterator.hasNext();
+        }
+
+        @Override
+        public A next() {
+            if (this.hasNext()) {
+                return this.action.transform(this.iterator.next());
+            } else {
```
yegor256: see http://www.yegor256.com/2015/01/21/if-then-throw-else.html

src/main/java/org/takes/misc/Transform.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        @Override
+        public boolean hasNext() {
+            return this.iterator.hasNext();
+        }
+
+        @Override
+        public A next() {
+            if (this.hasNext()) {
```
yegor256: I don't understand why do we need to call `hasNext()` here? we can just use `this.iterator.next()` and let it throw if there is nothing left

yegor256:
@super132 a few more comments above.. if you want to keep `LowerCase` inside `Condition`, you should make sure it does one thing at a time. Now this class is doing too many things. I would suggest to use `Transform.ToString` and `Select` in combination with `Condition.StartsWith`.

src/main/java/org/takes/misc/Select.java
super132: @yegor256, this is to initialise the first element for ```hasNext()``` and ```next()``` to work with and it is needed as part of iterator initialisation even though no one called ```.next()```

yegor256: @super132 you're calling `next()` at the next line

super132: @yegor256, I mean I need to obtain the first element, if any, for ```hasNext()``` and ```next()``` implementation to work on. I will think of a way to get rid of this first call in ctor. Thanks.

src/main/java/org/takes/misc/Concat.java
super132: @yegor256, thanks. I will modify this to avoid unnecessary call.

super132:
@yegor256, I have thought of it for sometime and I think that using ```Transform``` does not work as it will modify the content of the iterable while we are just selecting elements from that iterable with comparison to a prefix in lower case only. Without compound condition like in LowerCase now, I cannot think of a way to do it just by using one condition at a time. LowerCase basically is a complex StartsWith condition with element must be in lower case first. Can it be treated as "doing one thing at a time"? I could rename LowerCase to LowerStartsWith to avoid confusion.

I have also tried to move this out as a anonymous class but again, qulice complains about it before too long (22 lines).

super132:
@yegor256, another solution I can think of is to move the class ```LowerStartsWith``` as an inner class in the caller. This will then avoid the confusion of being a general condition that can be used by others. May I have your view on it?

yegor256:
`Transform` and `Select` should work in combination with each other. your class `LowerCase` is combining them inside, that's wrong. First you should transform and then select, or the other way around, but using different implementations. Did you ever use Guava constructs for that?

super132:
@yegor256, The original ```LowerCase``` class (now renamed to ```LowerStartsWith```) does not involve transformation. Transformation here is to changing all elements in original iterable. However, in this case, we are just comparing elements using lower case and it does not affect the output. For example, if the element in iterable is "Location: localhost" and if it passes the condition, the result iterable should contain "Location: localhost" but not "location: localhost". Therefore, I think using Transformation is inappropriate here. I have extracted the negation out as a general Condition here.

I will try look for Guava library to see if it fits the case. Thanks.

super132:
@yegor256, instead of using ```Transform```, I modifed ```LowerCase``` to be a decorative ```Condition``` which apply the supplied ```Condition``` by changing the element into lower case. ```StartsWith``` condition is created for checking prefix. I think in that way, we can have every Condition do one single action. 
I have also refactored ```SelectIterator``` to extends Guava ```AbstractIterator``` to have more clean implementation. Would you please review? Thanks.

pom.xml
```diff
@@ -136,6 +136,11 @@
             <optional>true</optional>
         </dependency>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <scope>compile</scope>
```
yegor256: we can't use any dependencies in this project

src/main/java/org/takes/misc/Condition.java
```diff
@@ -0,0 +1,119 @@
...
+        public boolean fits(final T element) {
+            return !this.condition.fits(element);
+        }
+    }
+
+    /**
+     * Translate the string element into lower case for condition checking.
+     * It does not alter the element in an iterable.
+     */
+    class LowerCase implements Condition<String> {
```
yegor256: condition can't be "lower case". case lowering is a transformation, not an operator.

yegor256:
@super132 again all wrong :) I would suggest you to first draw the design in text and suggest it. we'll discuss. then you will create Java code. looks like you don't understand the idea of functional programming and its application in OOP. I asked you whether you used Guava just to check whether you have any experience in using their "functional" concepts. besides that, your branch is very big in size, I would recommend to make it shorter and use PDD: http://www.yegor256.com/2009/03/04/pdd.html

pom.xml
super132: No problem. I can revert back to previous version.

super132:
@yegor256, I thought it would be ok this time :P I still do not think that the lower case here is a transformation. Let's take a look of the original codes in ```RqWithoutHeader```:
```java
for (final String header : req.head()) {
    if (!header.toLowerCase(Locale.ENGLISH)
        .startsWith(prefix)) {
        head.add(header);
    }
}
```
The logic here is to compare if header does not contain a prefix in a case insensitive manner and output it. And once it passes the condition, **the original header**, not the one transformed into lower case, is added to the result. Same logic here in the current implementation that I use a ```Condition``` to do this comparison by translating the header to lower case for case insensitive comparison. I do not see why do you think it is a transformation. Would you please point me to that? Would it be better to rename it to ```StartsWithIgnoreCase``` instead of ```LowerCase``` ?

If we had to implement it using transformation, then we have to do something like this:
```java
new Transform<String, String>(
    new Select<String>(
        new Transform<String, String>(
            req.head(),
            new TransformAction.LowerCase()
        ),
        new Not<String>(
            new Condition.StartsWith(prefix)
        )
    ),
    new TransformAction.NormalCase()
)
```
The problems in the above implementation are:
1. The implementation of ```NormalCase``` as ```LowerCase``` transformation is destructive and cannot be reverted unless we save a copy first and compare the transform result. But it would then defeat the purpose of this design being very light-weight and do not materialize iterable except the original input.
2. The above transformation would be slower than doing the case insensitive comparison in one ```Condition``` as it involves 2 transformations and one select. In my current implementation it involves one select only.

Based on the reasons above, I cannot see this comparison is a transformation, but a case insensitive comparison. If we really needed to implement it in transformation, would you please advise on how to implement transformation that can revert an iterable in lower case to its original case? Thanks.

src/main/java/org/takes/misc/Condition.java
super132: @yegor256, please advise on my comments below to see if it is more appropriate to name the condition to be "StartsWithIgnoreCase" instead of this. But again, this is not a transformation in the use case.

super132:
@yegor256, I am happy to create a puzzle for the case insensitive transformation implementaiton. Should I modify my codes based on my comments above (grouping the use of `toLowerCase()` and `startsWith()` in one condition) and make it a puzzle instead? Thanks.

yegor256:
@super132 don't worry about qulice complaining about too long anonymous class. You can just suppress this warning, as it's done, for example in `TkFallback`, like 108.

I would do it like this:

```java
new Select<String>(
  req.head(),
  new Condition<String>() {
    @Override
    public boolean fits(String header) {
      return header.toLowerCase().startsWith(prefix);
    }
  }
)
```

super132:
@yegor256 Thanks for your advice. I have moved the condition to be inline. Please have a check.

super132:
@yegor256, may I know if the latest change is fine for you? Thanks.

super132:
@yegor256 , may I have your opinion on it? It has been pending for 5 days. Thanks.

yegor256:
@super132 I'll update you tomorrow, sorry for the delay

yegor256:
@super132 take a look: http://www.yegor256.com/2015/04/30/iterating-adapter.html

super132:
@yegor256 , do you mean that I need to implement the Iterator to take in the data source directly instead of the iterator and copy the whole element into a buffer instead? So the ConcatIterator for example will become something like below?

```java
class ConcatIterator<E> implements Iterator<E> {
    private final Iterable<E> aitr;
    private final Iterable<E> bitr;
    private final Queue<E> buffer = new LinkedList<E>();
    public ConcatIterator(final Iterable<E> alist, final Iterable<E> blist) {
        this.aitr = alist;
        this.bitr = blist;
    }
    public boolean hasNext() {
        if (this.buffer.isEmpty()) {
            //iterate over aitr and bitr and save the item into queue
        }
        return !this.buffer.isEmpty();
    }
    public E next() {
         if (!this.hasNext()) {
             throw new NoSuchElementException("No element");
         }
         return this.buffer.poll();
    }
}
```

But then will it violate the rule you told me before that the iterable and the iterator should be as lazy as possible? The implementation now will do the actual condition check and transformation at the very last moment and using the implementation above, it would not be this lazy and it will scan through all elements even though the caller gets the first element in the end. Or do you want to change the implementation to take in the `Iterable` instead instead of the `Iterator` directly?

Another question I have about the implementation of `hasNext()`. If we use the approach above, should I get a new iterator in `hasNext()` method for the `Iterable` we saved? But then it would never stop as `Iterable.iterator()` method always return an `Iterator` with index set at zero, which will then loop through the whole list again. If we are going to save the `Iterator` on the instantiation of this object then it would be no different from the current implementation except taking the `Iterable`s as the constructor input and copy all the elements over from the `Iterable`s. 

With the questions above, may I confirm with you that would you want to have the implementation like the code block above? Thanks.

yegor256:
@super132 we're making this pull request too long. You're asking too many questions and I have to spend too much time to educate you. I would suggest to split this pull request into smaller parts. This one is too big and has too many issues.

super132:
@yegor256 , no problem. I will create puzzle on the Iterator design in all 3 iterator that I made and close this one. Would it be better to handle the issues like this? I'm going to do add the puzzle comments if you have no objection. Thanks.

yegor256:
@super132 in general, we prefer short branches

super132:
@yegor256 , let me split this PR into 3 smaller PRs, which will contain `Concat`, `Select` and `Transform` respectively and their affected classes.

super132:
@yegor256 , I have split this PR into 3 smaller PRs (#232, #233 and #234). Shall we close this PR and discuss in the child PRs? Thanks.

yegor256:
@super132 yes, I think this one should be closed

davvd:
@elenavolokhova review this ticket please, for compliance with our [quality rules](http://www.teamed.io/qa.html)

elenavolokhova:
@davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks for the review, we'll try better next time

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks a lot, next time everybody should try to make it better

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks for the review, we'll try better next time

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks a lot, next time everybody should try to make it better

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks, got it. everybody, please try to make it good next time

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks a lot, next time everybody should try to make it better

davvd:
> @davvd This request wasn't merged but performer should be rewarded for review. Quality is acceptable.

@elenavolokhova thanks, got it. everybody, please try to make it good next time

davvd:
@krzyk just added 10 mins to @elenavolokhova (for QA), payment ID is `56851691`... Much appreciated! **21 mins** was added to your account, payment ID `56851703`, time spent is 334 hours and 58 mins... you're getting extra minutes here (c=6)... +21 added to your rating, current score is: [+4914](http://www.netbout.com/b/35035?open=rating)

Issue #94, opened 2015-03-30
BkTimeable


yegor256:
Let's create a new decorator of `Back`, which will allow any particular call to `accept()` last for less than X milliseconds. Something like this:

```java
new BkTimeable(new BkParallel(new BkBasic(takes)), 5000);
```

I think we should start "watching" threads for each call to `accept()` and terminate them when time is running out.

Also, let's add a new argument to `FtCLI`: `--max-latency` to configure this behavior from command line

davvd:
@yegor256 OK, I labeled it as `bug`

davvd:
@eg04lt3r This task is yours, please go ahead keeping in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). If any questions, don't hesitate to ask right here

The budget of this task is **30 mins**. This is exactly how much time will be compensated, when the task is completed. Read about our [Definition of Done](http://www.xdsd.org/2014/04/15/definition-of-done.html)

davvd:
@yegor256 thanks for this report, I added 15 mins to your account, in transaction `000-882bb1bd`

eg04lt3r:
@yegor256 fixed in #123 pull request

davvd:
@eg04lt3r the ticket is with you for 15 days already. If it is not finished in the next 48 hours, it will be re-assigned to someone else, see [No Obligations](http://www.xdsd.org/2014/04/13/no-obligations-principle.html) principle, [this article](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html) should help if you're stuck... -30 added to your rating, current score is: [+8](http://www.netbout.com/b/35234?open=rating)

davvd:
@eg04lt3r it takes too long. Usually, we expect any task to be finished in less than a week. I'll assign someone else. Please stop working with it right now. See our [no obligations](http://www.xdsd.org/2014/04/13/no-obligations-principle.html) principle

-60 added to your rating, at the moment it is: [-82](http://www.netbout.com/b/35234?open=rating)

davvd:
@dmzaytsev the task is yours please proceed

dmzaytsev:
@davvd I created the PR #197  for this task. Could you please find someone to review ?

dmzaytsev:
@yegor256 the PR #197 merged. Please close the task if alright

yegor256:
@dmzaytsev thanks a lot!

yegor256:
@rultor release, tag is `0.15`

[...]

PR #197, opened 2015-04-21
#94 BkTimeable

dmzaytsev:
For #94 

* created BkTimeable -  Back decorator with maximum lifetime
* created unit test for BkTimeable
* added  a new argument to FtCLI to configure this behavior from command line

### DIFF
--- a/src/main/java/org/takes/http/BkTimeable.java
+++ b/src/main/java/org/takes/http/BkTimeable.java
@@ -1,0 +1,131 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.takes.http;
+
+import java.io.IOException;
+import java.net.Socket;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import lombok.EqualsAndHashCode;
+
+/**
+ * Back decorator with maximum lifetime.
+ *
+ * <p>The class is immutable and thread-safe.
+ * @author Dmitry Zaytsev (dmitry.zaytsev@gmail.com)
+ * @version $Id$
+ * @since 0.14.2
+ */
+@EqualsAndHashCode(of = {"origin", "latency" })
+public final class BkTimeable implements Back {
+    /**
+     * Original back.
+     */
+    private final transient Back origin;
+
+    /**
+     * Maximum latency in milliseconds.
+     */
+    private final transient long latency;
+
+    /**
+     * Threads storage.
+     */
+    private final transient CopyOnWriteArraySet<BkTimeable.ThreadInfo> threads;
+
+    /**
+     * Monitoring executor.
+     */
+    private final transient ScheduledExecutorService service;
+
+    /**
+     * Ctor.
+     * @param back Original back
+     * @param msec Execution latency
+     */
+    public BkTimeable(final Back back, final long msec) {
+        this.origin = back;
+        this.latency = msec;
+        this.threads = new CopyOnWriteArraySet<BkTimeable.ThreadInfo>();
+        this.service = Executors.newSingleThreadScheduledExecutor();
+        this.start();
+    }
+
+    @Override
+    public void accept(final Socket socket) throws IOException {
+        this.threads.add(new BkTimeable.ThreadInfo(Thread.currentThread()));
+        this.origin.accept(socket);
+    }
+
+    /**
+     * Start monitoring.
+     */
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    private void start() {
+        final long time = this.latency;
+        final CopyOnWriteArraySet<BkTimeable.ThreadInfo> storage = this.threads;
+        this.service.scheduleAtFixedRate(
+            new Runnable() {
+                @Override
+                public void run() {
+                    for (final BkTimeable.ThreadInfo info : storage) {
+                        if (System.currentTimeMillis() - info.start > time) {
+                            info.thread.interrupt();
+                            storage.remove(info);
+                        }
+                    }
+                }
+            },
+            0,
+            1,
+            TimeUnit.MILLISECONDS
+        );
+    }
+
+    /**
+     * Thread info.
+     */
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    public static final class ThreadInfo {
+        /**
+         * Thread reference.
+         */
+        private final transient Thread thread;
+        /**
+         * Start time.
+         */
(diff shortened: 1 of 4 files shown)


davvd:
@dmzaytsev Thanks, let me find someone who can review this pull request

davvd:
@ggajos this pull request is for you, please review

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import lombok.EqualsAndHashCode;
+
+/**
+ * Back decorator with maximum lifetime.
+ *
+ * <p>The class is immutable and thread-safe.
```
ggajos: @dmzaytsev you might want to close <p> tag

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,138 @@
...
+ * @author Dmitry Zaytsev (dmitry.zaytsev@gmail.com)
+ * @version $Id$
+ * @since 0.14.2
+ */
+@EqualsAndHashCode(of = {"origin", "latency" })
+public final class BkTimeable implements Back {
+    /**
+     * Threads check interval.
+     */
+    private static final long CHECK_INTERVAL = 100;
```
ggajos: @dmzaytsev think we should avoid extracting variables that are used only once

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,138 @@
...
+        this.threads.add(new BkTimeable.ThreadInfo(Thread.currentThread()));
+        this.origin.accept(socket);
+    }
+
+    /**
+     * Start monitoring.
+     */
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    private void start() {
+        final long time = this.latency;
```
ggajos: @dmzaytsev we should avoid extracting to local final variable if used only once

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Maximum latency in milliseconds.
+     */
+    private final transient long latency;
+
+    /**
+     * Threads storage.
+     */
+    private final transient CopyOnWriteArraySet<BkTimeable.ThreadInfo> threads;
```
ggajos: @dmzaytsev think we should use interface here

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,138 @@
...
+        this.origin = back;
+        this.latency = msec;
+        this.threads = new CopyOnWriteArraySet<BkTimeable.ThreadInfo>();
+        this.service = Executors.newSingleThreadScheduledExecutor();
+        this.start();
+    }
+
+    @Override
+    public void accept(final Socket socket) throws IOException {
+        this.threads.add(new BkTimeable.ThreadInfo(Thread.currentThread()));
```
ggajos: @dmzaytsev ThreadInfo can read currentThread in Ctor, no need to pass static

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Ctor.
+     * @param back Original back
+     * @param msec Execution latency
+     */
+    public BkTimeable(final Back back, final long msec) {
+        this.origin = back;
+        this.latency = msec;
+        this.threads = new CopyOnWriteArraySet<BkTimeable.ThreadInfo>();
+        this.service = Executors.newSingleThreadScheduledExecutor();
```
ggajos: @dmzaytsev we should avoid creating executor service here, if we create additional Ctor where we can pass executor service then we're able also to mock this service and is much simpler for testing; otherwise the only way to test is to use some sleep and latches

src/test/java/org/takes/http/BkTimeableTest.java
```diff
@@ -0,0 +1,88 @@
...
+        final Back back = new BkTimeable(
+            new Back() {
+                @Override
+                public void accept(final Socket socket) throws IOException {
+                    new Runnable() {
+                        @Override
+                        public void run() {
+                            final long start = System.currentTimeMillis();
+                            try {
+                                TimeUnit.MILLISECONDS.sleep(1000);
```
ggajos: @dmz we should avoid using sleep

ggajos:
@dmzaytsev please see comments above

src/test/java/org/takes/http/BkTimeableTest.java
dmzaytsev: @ggajos I simulate long running process here.
Could you suggest any acceptable replacement for Threads.sleep ?

src/main/java/org/takes/http/BkTimeable.java
dmzaytsev: @ggajos this.latency not visible for the inline Runnable below

dmzaytsev:
@ggajos thanks for your comments. 
I fixed some remarks, but I have a couple questions above.

src/main/java/org/takes/http/BkTimeable.java
dmzaytsev: @ggajos do you mean  `<p>The class is immutable and thread-safe.</p>` ?
but every class in takes have no  `</p>` tag

src/test/java/org/takes/http/BkTimeableTest.java
ggajos: @dmzaytsev you should not rely on this based on test, you should use latches or mock some behaviour

dmzaytsev: @ggajos BkTimeable is a watcher. Its purpose to interrupt long running process.
I don't use `Thread.sleep()` to test BkTimeable. I simulate a long-running process via `Thread.sleep()`  call. Unfortunally I do not quite understand how should I mock such process.

I can replace `Thread.sleep(1000)` to `new CountDownLatch(1).await(1000, TimeUnit.MILLISECONDS)` but I think it will be less readable. In addition, it will throw other InterruptedException. 

What's wrong with `Thread.sleep()` in this case ?

ggajos: @dmzaytsev sleep in testing is
* just bad practice
* you can never be sure that that you're sleeping long enough
* you're relying on implementation of something else, in this particular case you're testing scheduled executor service

I would advice to pass mocked executor service to BkTimeable which will call you're runnable immidiately. Think that this test will be much simpler if you can control also scheduled service. What do you think?

dmzaytsev: @ggajos I think this test won't be simpler. Suppose I pass mocked executor BkTimeable, but I still need to use different threads in this test. Because  BkTimeable.accept() use current thread to add it to the check list.
but I can try...

dmzaytsev:
@ggajos look at the changes please

src/main/java/org/takes/http/BkTimeable.java
ggajos: @dmzaytsev you're right, javadoc convention doesn't require to close `</p>` tag

ggajos:
@dmzaytsev thanks! there is no sleep in tests, cool :)

ggajos:
@rultor merge please

dmzaytsev:
@ggajos thanks for review! :)

ggajos:
@dmz welcome :)

src/main/java/org/takes/http/BkTimeable.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public void accept(final Socket socket) throws IOException {
+        this.threads.add(new BkTimeable.ThreadInfo());
+        this.origin.accept(socket);
+    }
+
+    /**
+     * Start monitoring.
+     * @checkstyle MagicNumberCheck (25 lines)
```
yegor256: I would place this right before `100`. now you're suppressing 25 lines, it's not healthy

src/test/java/org/takes/http/BkTimeableTest.java
```diff
@@ -0,0 +1,106 @@
...
+            threads
+        );
+        back.accept(new Socket());
+        MatcherAssert.assertThat(threads, Matchers.hasSize(1));
+    }
+
+    /**
+     * BkTimeable can stop running thread.
+     * @throws java.io.IOException If some problem inside
+     * @checkstyle MagicNumberCheck (500 lines)
```
yegor256: I don't think you need this at all

yegor256:
@dmzaytsev looks good, thanks! just a few small comments

dmzaytsev:
@yegor256 I removed not healthy checkstyle suppression
take a look please

yegor256:
@rultor merge pls

davvd:

davvd:
@rultor deploy now pls

elenavolokhova:
@davvd This ticket looks good. Thank you.

davvd:
> @davvd This ticket looks good. Thank you.

@elenavolokhova thanks a lot

davvd:
@ggajos just added 10 mins to @elenavolokhova (for QA), payment ID is `56385932`; Thanks so much! Your account was topped up for **19 mins** (transaction ID is `AP-4AM89881H9432581D`, the task took 70 hours and 50 mins); review comments (c=4) added as a bonus; +19 added to your rating, current score is: [+3338](http://www.netbout.com/b/35094?open=rating)

Issue #130, opened 2015-04-05
socket information is not present in request


yegor256:
When `Request` gets into a take, it doesn't have any information about the socket it came from. We're missing a lot of useful information, which is present in Java servlets, for example, like IP address of the client, etc.

Let's add extra custom headers to each request, inside `BkBasic`. Something like `X-Takes-LocalAddress`, `X-Takes-LocalPort`, etc. As much information as we can get from a `Socket`.

Then, let's create a class `RqSocket`, which will help to extract that information from the request, like:

```java
final class RqSocket extends RqWrap {
  InetAddress localAddress();
  int localPort();
  InetAddress remoteAddress();
  int remotePort();
  // etc.
}
```

davvd:
@yegor256 got it, tag `bug` is here

davvd:
@d-molotchko please pick this up, and keep in mind [these instructions](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Any technical questions - ask right here

This task's budget is **30 mins**. This is exactly how much will be paid when the problem explained above is solved. See [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for more information

davvd:
@yegor256 thanks a lot for reporting, **15 mins** added to your acc, pmt ID `000-a435c29b`

d-molotchko:
@davvd please reassign this task to another dev. thanks.

yegor256:
@davvd give this to somebody else

davvd:
> @davvd give this to somebody else

@yegor256 no problem, I will try to find somebody else

davvd:
@yegor256 milestone set to `1.0` (correct me if I am wrong)

davvd:
@cyberone can you please help? Keep in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). If you have any technical questions, don't hesitate to ask right here

Total fixed cost of this task is **30 mins** (see [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for more info)

cyberone:
@davvd assign somebody else please

davvd:
> @davvd assign somebody else please

@cyberone 30 points deducted from your rating

davvd:
> @davvd assign somebody else please

@cyberone right, I will find someone else, no problem

davvd:
@ikhvostenkov it's yours,please go ahead

carlosmiranda:
[...]

PR #225, opened 2015-04-29
 Socket information in request

ikhvostenkov:
This is fix for https://github.com/yegor256/takes/issues/130
Added ```RqSocket``` decorator to extract socket specific information from the request. 
Edited ```BkBasic``` to inject socket specific information into request.

### DIFF
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -129,14 +129,19 @@ public final class BkBasic implements Back {
     private static Request addSocketHeaders(final Request req,
         final Socket socket) {
         return new RqWithHeaders(
-            req, new StringBuilder().append("X-Takes-LocalAddress: ")
-            .append(socket.getLocalAddress().toString())
-            .append("X-Takes-LocalPort")
-            .append(Integer.toString(socket.getLocalPort()))
-            .append("X-Takes-RemoteAddress")
-            .append(socket.getInetAddress().toString())
-            .append("X-Takes-RemotePort")
-            .append(Integer.toString(socket.getPort()))
+            req,
+            String.format(
+                "X-Takes-LocalAddress: %s", socket.getLocalAddress().toString()
+            ),
+            String.format(
+                "X-Takes-LocalPort: %d", socket.getLocalPort()
+            ),
+            String.format(
+                "X-Takes-RemoteAddress: %s", socket.getInetAddress().toString()
+            ),
+            String.format(
+                "X-Takes-RemotePort: %d", socket.getPort()
+            )
         );
     }
 }
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -38,6 +38,7 @@ import org.takes.Request;
 import org.takes.Response;
 import org.takes.Take;
 import org.takes.rq.RqLive;
+import org.takes.rq.RqWithHeader;
 import org.takes.rs.RsPrint;
 import org.takes.rs.RsText;
 import org.takes.rs.RsWithStatus;
@@ -73,7 +74,10 @@ public final class BkBasic implements Back {
         final InputStream input = socket.getInputStream();
         try {
             this.print(
-                new RqLive(new BufferedInputStream(input)),
+                BkBasic.addSocketHeaders(
+                    new RqLive(new BufferedInputStream(input)),
+                    socket
+                ),
                 new BufferedOutputStream(socket.getOutputStream())
             );
         } finally {
@@ -114,6 +118,34 @@ public final class BkBasic implements Back {
             new RsText(new ByteArrayInputStream(baos.toByteArray())),
             HttpURLConnection.HTTP_INTERNAL_ERROR
         );
+    }
+
+    /**
+     * Adds custom headers with information about socket.
+     * @param req Request
+     * @param socket Socket
+     * @return Request with custom headers
+     */
+    private static Request addSocketHeaders(final Request req,
+        final Socket socket) {
+        Request mreq = req;
+        mreq = new RqWithHeader(
+            mreq, "X-Takes-LocalAddress",
+            socket.getLocalAddress().toString()
+        );
+        mreq = new RqWithHeader(
+            mreq, "X-Takes-LocalPort",
+            Integer.toString(socket.getLocalPort())
+        );
+        mreq = new RqWithHeader(
+            mreq, "X-Takes-RemoteAddress",
+            socket.getInetAddress().toString()
+        );
+        mreq = new RqWithHeader(
+            mreq, "X-Takes-RemotePort",
+            Integer.toString(socket.getPort())
+        );
+        return mreq;
     }
 
 }
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -37,7 +37,7 @@ import org.takes.Request;
 import org.takes.Response;
 import org.takes.Take;
 import org.takes.rq.RqLive;
-import org.takes.rq.RqWithHeader;
+import org.takes.rq.RqWithHeaders;
 import org.takes.rs.RsPrint;
 import org.takes.rs.RsText;
 import org.takes.rs.RsWithStatus;
@@ -128,15 +128,15 @@ public final class BkBasic implements Back {
      */
     private static Request addSocketHeaders(final Request req,
         final Socket socket) {
-        return new RqWithHeader(
-            req, "X-Takes-LocalAddress", socket.getLocalAddress().toString()
-        ).withAdditionalHeader(
-            "X-Takes-LocalPort", Integer.toString(socket.getLocalPort())
-        ).withAdditionalHeader(
-            "X-Takes-RemoteAddress", socket.getInetAddress().toString()
-        ).withAdditionalHeader(
-            "X-Takes-RemotePort", Integer.toString(socket.getPort())
+        return new RqWithHeaders(
+            req, new StringBuilder().append("X-Takes-LocalAddress: ")
+            .append(socket.getLocalAddress().toString())
+            .append("X-Takes-LocalPort")
+            .append(Integer.toString(socket.getLocalPort()))
+            .append("X-Takes-RemoteAddress")
+            .append(socket.getInetAddress().toString())
+            .append("X-Takes-RemotePort")
+            .append(Integer.toString(socket.getPort()))
         );
     }
-
 }
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -37,6 +37,7 @@ import org.takes.Request;
 import org.takes.Response;
 import org.takes.Take;
 import org.takes.rq.RqLive;
+import org.takes.rq.RqWithHeader;
 import org.takes.rs.RsPrint;
 import org.takes.rs.RsText;
 import org.takes.rs.RsWithStatus;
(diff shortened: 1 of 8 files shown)


ikhvostenkov:
@davvd PR is ready for review.

davvd:
@ikhvostenkov Many thanks for the PR, let me find a reviewer for it

davvd:
@carlosmiranda review this one, please

davvd:
> @davvd PR is ready for review.

@ikhvostenkov will take a look, thanks

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -116,4 +120,32 @@ private static Response failure(final Throwable err) {
...
+    /**
+     * Adds custom headers with information about socket.
+     * @param req Request
+     * @param socket Socket
+     * @return Request with custom headers
+     */
+    private static Request addSocketHeaders(final Request req,
+        final Socket socket) {
+        Request mreq = req;
+        mreq = new RqWithHeader(
```
carlosmiranda: @ikhvostenkov I don't like this design. How about adding a method `withAdditionalHeader` to `RqWithHeader` that simply returns `new RqWithHeader(this, "Header-Name", "Header-Value")`. That way we can get a request with multiple headers through a fluent interface like this:

```java
return new RqWithHeader(mreq, "Header-1", "Value-1")
    .withAdditionalHeader("Header-2", "value-2")
    .withAdditionalHeader("Header-3", "value-3")
```

src/main/java/org/takes/rq/RqSocket.java
```diff
@@ -0,0 +1,119 @@
...
+        return result;
+    }
+
+    /**
+     * Parses port value from the provided header.
+     * @param header Header to parse
+     * @return Integer value of the port
+     * @throws IOException If fails
+     */
+    private int getPort(final CharSequence header) throws IOException {
```
carlosmiranda: @ikhvostenkov I don't see the value of this method exactly. Why are we defaulting to zero if there is no header value? It should throw an exception just like in other cases. Let's just call `Integer.parseInt(this.parseHeader(header))`.

src/main/java/org/takes/rq/RqSocket.java
```diff
@@ -0,0 +1,119 @@
...
+        return this.getPort("X-Takes-RemotePort");
+    }
+
+    /**
+     * Parses value of the provided header from head.
+     * @param header Header to parse
+     * @return String value of the provided header
+     * @throws IOException If fails
+     */
+    private String parseHeader(final CharSequence header) throws IOException {
```
carlosmiranda: @ikhvostenkov It would be more accurate to call this method `headerValue`, since it only returns the value portion of the header.

carlosmiranda:
@ikhvostenkov Looks good, in general, I do have some comments above.

src/main/java/org/takes/http/BkBasic.java
ikhvostenkov: @carlosmiranda I agree with you, I do not like this design as well, my thoughts were to refactor ```RqWithHeader``` in such way, that we could add multiply headers in the same method using for example hash map. 
@yegor256 Could you please help here? What are your thoughts on how we shall proceed here?

carlosmiranda: @ikhvostenkov No need for `HashMap`. Let's keep the API immutable. How about adding a method like this:

```java
public RqWithHeader withAdditionalHeader(CharSequence name, CharSequence value) {
    return new RqWithHeader(this, name, value);
}
```

This should enable the fluent interface that I suggested.

carlosmiranda:
@ikhvostenkov Replied to your comment above. Also please do a resync with master, there are merge conflicts.

ikhvostenkov:
@carlosmiranda Thank you for your comments. I fixed all issues above.

src/test/java/org/takes/rq/RqWithHeaderTest.java
```diff
@@ -53,4 +53,24 @@ public void addsHttpHeaders() throws IOException {
         );
     }
 
+    /**
+     * RqWithHeader can add multiply additional headers.
```
carlosmiranda: @ikhvostenkov "multiple" instead of "multiply"

src/test/java/org/takes/rq/RqWithHeaderTest.java
```diff
@@ -53,4 +53,24 @@ public void addsHttpHeaders() throws IOException {
...
+     * @throws IOException If some problem inside
+     */
+    @Test
+    public void addsMultiplyHeaders() throws IOException {
```
carlosmiranda: @ikhvostenkov same here... should be named `addsMultipleHeaders`

src/main/java/org/takes/rq/RqWithHeader.java
```diff
@@ -78,4 +78,15 @@ public InputStream body() throws IOException {
         );
     }
 
+    /**
+     * Add additional multiply headers.
```
carlosmiranda: @ikhvostenkov We're only adding one header here, so it's better to simply say "Add additional header".

src/main/java/org/takes/rq/RqSocket.java
carlosmiranda: @ikhvostenkov Following you up on my previous comment.

carlosmiranda:
@ikhvostenkov Just a few more comments.

ikhvostenkov:
@carlosmiranda fixed errors above.

src/main/java/org/takes/rq/RqSocket.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Parses port value from the provided header.
+     * @param header Header to parse
+     * @return Integer value of the port
+     * @throws IOException If fails
+     */
+    private int getPort(final CharSequence header) throws IOException {
+        return Integer.parseInt(this.headerValue(header));
```
carlosmiranda: @ikhvostenkov No use for having this method now. Let's remove it and inline `Integer.parseInt(...)` call in the methods `getLocalPort` and `getRemotePort` above.

carlosmiranda:
@ikhvostenkov Just one more comment above and then we're good.

src/main/java/org/takes/rq/RqSocket.java
ikhvostenkov: @carlosmiranda try, method is redundant.

carlosmiranda:
@rultor it's time to merge

src/main/java/org/takes/rq/RqSocket.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Parses value of the provided header from head.
+     * @param header Header to parse
+     * @return String value of the provided header
+     * @throws IOException If fails
+     */
+    private String headerValue(final CharSequence header) throws IOException {
+        String result = "";
+        for (final String line : this.head()) {
```
yegor256: looks like you're duplicating here what we already have in `RqHeaders`

src/main/java/org/takes/rq/RqWithHeader.java
```diff
@@ -78,4 +78,15 @@ public InputStream body() throws IOException {
         );
     }
 
+    /**
+     * Add additional header.
+     * @param name Header name
+     * @param value Header value
+     * @return Request with extra header
+     */
+    public RqWithHeader withAdditionalHeader(final CharSequence name,
```
yegor256: I don't think it's a good idea. We don't have this methods anywhere. We're always using constructors. Let's follow this convention. You can create `RqWithHeaders`, similar to `RsWithHeaders`, if you want to make that cascading object constructing look "shorter"

yegor256:
@ikhvostenkov two comments from me above

ikhvostenkov:
@yegor256 fixed all comments from your site.

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -115,4 +120,23 @@ private static Response failure(final Throwable err) {
...
+        final Socket socket) {
+        return new RqWithHeaders(
+            req, new StringBuilder().append("X-Takes-LocalAddress: ")
+            .append(socket.getLocalAddress().toString())
+            .append("X-Takes-LocalPort")
+            .append(Integer.toString(socket.getLocalPort()))
+            .append("X-Takes-RemoteAddress")
+            .append(socket.getInetAddress().toString())
+            .append("X-Takes-RemotePort")
+            .append(Integer.toString(socket.getPort()))
```
yegor256: brrr... are you sure that how it should be? you're concatenating them all into one string... I think it should be something like:

```java
new RqWithHeaders(
  req,
  String.format("X-Takes-LocalAddress: %s", socket.getLocalAddress()),
  String.format("X-Takes-LocalPort: %d", socket.getLocalPort()),
  // etc.
)
```

yegor256:
@ikhvostenkov see above

ikhvostenkov:
@yegor256 thank you for catching this error, do not know what was thinking about.

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -115,4 +120,28 @@ private static Response failure(final Throwable err) {
...
+     * @param req Request
+     * @param socket Socket
+     * @return Request with custom headers
+     */
+    private static Request addSocketHeaders(final Request req,
+        final Socket socket) {
+        return new RqWithHeaders(
+            req,
+            String.format(
+                "X-Takes-LocalAddress: %s", socket.getLocalAddress().toString()
```
yegor256: `toString` is redundant here, `String.format()` calls it by default

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -115,4 +120,28 @@ private static Response failure(final Throwable err) {
...
+            String.format(
+                "X-Takes-LocalPort: %d", socket.getLocalPort()
+            ),
+            String.format(
+                "X-Takes-RemoteAddress: %s", socket.getInetAddress().toString()
+            ),
+            String.format(
+                "X-Takes-RemotePort: %d", socket.getPort()
+            )
+        );
```
yegor256: if you remove `toString` you can format it better, keeping one header per line

src/main/java/org/takes/rq/RqSocket.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * Returns IP address from the X-Takes-LocalAddress header.
+     * @return Local InetAddress
+     * @throws IOException If fails
+     */
+    public InetAddress getLocalAddress() throws IOException {
+        return InetAddress.getByName(
+            new RqHeaders.Base(this).header("X-Takes-LocalAddress")
```
yegor256: I think it's better to use `RqHeaders.Smart#single()` here (and all methods below)

yegor256:
@ikhvostenkov two more comments

ikhvostenkov:
@yegor256 thank you for your comments. Issues above have been fixed.

yegor256:
@rultor try to merge

davvd:
@elenavolokhova please, review this ticket for compliance with our [QA rules](http://www.teamed.io/qa.html)

elenavolokhova:
@davvd This PR looks good.

davvd:
> @davvd This PR looks good.

@elenavolokhova thank you

davvd:
@carlosmiranda just added 10 mins to @elenavolokhova (for QA), payment ID is `56894534`

Much obliged! I have added **20 mins** to your account in payment "56894543", 168 hours and 45 mins spent

review comments (c=5) added as a bonus

added +20 to your rating, now it is equal to [+4521](http://www.netbout.com/b/35023?open=rating)

davvd:
@rultor deploy now pls

Issue #239, opened 2015-05-01
PsLinkedin passes the request parameters to the URI.


dmzaytsev:
PsLinkedin passes the request parameters to the URI instead of the body.
```
POST /uas/oauth2/accessToken?grant_type=authorization_code&client_id=clnid&redirect_uri=http%3A%2F%2Fwww.example.com%2F&client_secret=secret&code=xxxx HTTP/1.1
Accept: application/xml
Host: localhost:50809
Connection: keep-alive
Content-type: application/x-www-form-urlencoded
Content-Length: 0
```
but should be 
```
POST /uas/oauth2/accessToken HTTP/1.1
Accept: application/xml
Host: localhost:50809
Connection: keep-alive
Content-type: application/x-www-form-urlencoded

grant_type=authorization_code&client_id=clnid&redirect_uri=http%3A%2F%2Fwww.example.com%2F&client_secret=secret&code=xxxx
```
Linkedin accepts such request but it is undocumented feature 
I think we should use valid POST requests

davvd:
@dmzaytsev this task will get someone's attention soon

davvd:
@dmzaytsev attached this issue to milestone "1.0" (let me know if this is wrong)

davvd:
@dmzaytsev many thanks for the report, I topped your account for 15 mins, transaction 56454744

davvd:
@super132 the task is for you now, follow [these guidelines](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Don't hesitate to ask any technical questions right here

The budget here is **30 mins**, which is exactly how much time will be paid for, when the task is [completed](http://www.xdsd.org/2014/04/15/definition-of-done.html)

super132:
@yegor256 PR #245 fixes this issue and merged successfully. Would you please close this ticket? Thanks.

yegor256:
@super132 thanks!

davvd:
@yegor256 only @dmzaytsev can close this issue, since he opened it

dmzaytsev:
@super132 thanks!

davvd:
@elenavolokhova please, review this ticket for compliance with our [QA rules](http://www.teamed.io/qa.html)

elenavolokhova:
@davvd Looks good!

davvd:
> @davvd Looks good!

@elenavolokhova thank you

davvd:
[...]

PR #244, opened 2015-05-02
#239 PsLinkedin passes the request parameters to the URI.

super132:
For task #239, removing the parameters in `Href` object and append them in `RequestBody`


super132:
@davvd , please review.

Issue #292, opened 2015-05-15
FbStatus without second argument


yegor256:
Let's create a new ctor in `FbStatus`, with a single `int` argument. When being used like this:

```java
new FbStatus(404);
```

It should return a response, which looks like this:

```
HTTP/1.1 404 Not Found
Content-Type: text/plain

404 Not Found: <exception message>
```

Where `<exception message>` is the message of the exception from `RqFallback`

davvd:
@yegor256 I tagged this as "bug"

davvd:
@yegor256 since there is no milestone yet I set it to "1.0"

davvd:
@komaz the task is for you now, follow [these guidelines](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Don't hesitate to ask any technical questions right here; Total fixed cost of this task is **30 mins** (see [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for more info)

davvd:
@yegor256 thank you for reporting this, I added 15 mins to your acc, transaction `000-dc409047`

komaz:
@davvd can I take this issue back?

yegor256:
@davvd assign @komaz to this task pls

davvd:
> @davvd assign @komaz to this task pls

@yegor256 no problem, @komaz is already assigned to this task, please go on

komaz:
@yegor256 the mapping from status code to explanation message is currently defind in private static map `RsWithStatus::REASONS`. Is it ok to extract this map into a separate class `Status`, so that both `RsWithStatus` and `FbStatus` would use `new Status(code).message()`? 

Actually I did all changes in [my branch](https://github.com/komaz/takes/commit/72bc42eda24786510ed62cee79d241d4d41a9b20), but as it depends on #295 fix, I'm not submitting a pull request yet.

yegor256:
@komaz I don't think it's a good idea. in `FbStatus` you have an instance of `Response` already. you can extract that status line from it. Right?

komaz:
@yegor256 I think if instantiate `FbStatus` only with `code`argument, all I have is `RqFallback` instance. Indeed, I can instantiate `RsWithStatus(code)`, take a `head()` from it, strip `HTTP 1.1 ` from the first line, and use the rest as the prefix for response body, but I thought it would be a leak of abstraction here and relying on the fact that `RsWithStatus` forms the first line of response head correctly. Am I missing something?

[...]

PR #303, opened 2015-05-19
#292 FbStatus without second argument

komaz:
Fix for issue #292. As discussed in issue, I'm taking a response code with explanation message from `RsWithStatus` head.

### DIFF
--- a/src/main/java/org/takes/facets/fallback/FbStatus.java
+++ b/src/main/java/org/takes/facets/fallback/FbStatus.java
@@ -32,6 +32,9 @@ import lombok.EqualsAndHashCode;
 import org.takes.Response;
 import org.takes.Take;
 import org.takes.misc.Condition;
+import org.takes.rs.RsWithBody;
+import org.takes.rs.RsWithStatus;
+import org.takes.rs.RsWithType;
 import org.takes.tk.TkFixed;
 
 /**
@@ -45,7 +48,46 @@ import org.takes.tk.TkFixed;
  */
 @EqualsAndHashCode(callSuper = true)
 public final class FbStatus extends FbWrap {
-
+    /**
+     * Ctor.
+     * @param code HTTP status code
+     * @since 0.16.10
+     */
+    public FbStatus(final int code) {
+        this(new Condition<Integer>() {
+            @Override
+            public boolean fits(final Integer status) {
+                return code == status.intValue();
+            }
+        });
+    }
+    /**
+     * Ctor.
+     * @param check HTTP status code predicate
+     * @since 0.16.10
+     */
+    public FbStatus(final Condition<Integer> check) {
+        this(check, new Fallback() {
+            @Override
+            public Iterator<Response> route(final RqFallback req)
+                throws IOException {
+                final Response res = new RsWithStatus(req.code());
+                return Collections.<Response>singleton(
+                    new RsWithType(
+                        new RsWithBody(
+                            res,
+                            String.format(
+                                "%s: %s",
+                                res.head().iterator().next().split("\\s", 2)[1],
+                                req.throwable().getLocalizedMessage()
+                            )
+                        ),
+                        "text/plain"
+                    )
+                ).iterator();
+            }
+        });
+    }
     /**
      * Ctor.
      * @param code HTTP status code
--- a/src/test/java/org/takes/facets/fallback/FbStatusTest.java
+++ b/src/test/java/org/takes/facets/fallback/FbStatusTest.java
@@ -23,6 +23,7 @@
  */
 package org.takes.facets.fallback;
 
+import java.io.IOException;
 import java.net.HttpURLConnection;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
@@ -34,7 +35,8 @@ import org.takes.tk.TkFixed;
 
 /**
  * Test case for {@link FbStatus}.
- * @author Ivan Inozemtsev (dmitry.zaytsev@gmail.com)
+ * @author Ivan Inozemtsev (ivan.inozemtsev@gmail.com)
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  * @version $Id$
  * @since 0.16.10
  */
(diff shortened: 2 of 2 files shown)


davvd:
@komaz I will find a reviewer for your pull requests shortly, thanks for contribution!

davvd:
@ggajos it's yours,please review

src/main/java/org/takes/facets/fallback/FbStatus.java
```diff
@@ -45,7 +47,46 @@
...
+     * @since 0.16.10
+     */
+    public FbStatus(final Condition<Integer> check) {
+        this(check, new Fallback() {
+            @Override
+            public Iterator<Response> route(final RqFallback req)
+                throws IOException {
+                final Response res = new RsWithStatus(req.code());
+                final String message = res
+                    .head().iterator().next().substring("HTTP 1.1 ".length());
```
ggajos: @komaz that's not looking good, what if HTTP will look like `HTTP/1.1 200 OK`?

src/test/java/org/takes/facets/fallback/FbStatusTest.java
```diff
@@ -102,4 +104,31 @@ public void ignoresDifferentStatus() throws Exception {
             Matchers.equalTo(false)
         );
     }
+
+    /**
+     * Default response body should contain status code,
```
ggajos: @komaz wrong comment, http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#test-method-names

src/main/java/org/takes/facets/fallback/FbStatus.java
```diff
@@ -45,7 +47,46 @@
...
+    /**
+     * Ctor.
+     * @param code HTTP status code
+     * @since 0.16.10
+     */
+    public FbStatus(final int code) {
+        this(new Condition<Integer>() {
+            @Override
+            public boolean fits(final Integer status) {
+                return code == status;
```
ggajos: @komaz autoboxing will be used here, but maybe let's make it visible? .`intValue()` or `equals()`

ggajos:
@komaz few comments

src/main/java/org/takes/facets/fallback/FbStatus.java
komaz: @ggajos in that case `200 OK` will be extracted, and the full message in the body will be something like `200 OK: <message from throwable>`. I assume that's exactly what we need, right? If someone creates fallback for status 200, then probably that's what he or she wants. Or do you mean there's a typo in a string, i.e. it should be `HTTP/1.1` instead of `HTTP 1.1`?

komaz:
@ggajos thanks for the review! I fixed test case javadoc, added `.intValue()` in condition and fixed a typo in `HTTP/1.1` prefix. Could you please comment what is wrong with extracting only a status and status message from a head?

ggajos:
@komaz I think that relying on version number is bad. Maybe we can use some regexp or `space` character so this implementation will still work when we deal with http/1.0. What do you think?

komaz:
@ggajos The code relies on a string length (`line.substring("HTTP/1.1 ".length())`), so I guess as long as this project won't respond with something like `HTTP/2.10`, or `HTTP/1.11`, it is safe. I'm afraid using regex like `HTTP/1\\.1\\s+(\\d+\\s?.*)` would make things even less readable and require dealing with matcher groups.

Maybe splitting this long expression to several lines would help? Something like this:
```java
final String first = res.head().iterator().next();
final int prefix = "HTTP/1.1 ".length();
final String message = first.substring("HTTP/1.1 ".length());
```
But in that case there'll be two local variables which are used only once

ggajos:
@komaz Current solution is ok I just wondering how we can avoid eventual surprises in future but maybe let's not over engineer this. Thanks for explanations.

ggajos:
@rultor merge please

src/main/java/org/takes/facets/fallback/FbStatus.java
```diff
@@ -45,7 +47,46 @@
...
+     * @param check HTTP status code predicate
+     * @since 0.16.10
+     */
+    public FbStatus(final Condition<Integer> check) {
+        this(check, new Fallback() {
+            @Override
+            public Iterator<Response> route(final RqFallback req)
+                throws IOException {
+                final Response res = new RsWithStatus(req.code());
+                final String message = res
```
yegor256: `message` looks like a one-time variable, see http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#one-time-variables

src/main/java/org/takes/facets/fallback/FbStatus.java
```diff
@@ -45,7 +47,46 @@
...
+                    .head().iterator().next().substring("HTTP/1.1 ".length());
```
yegor256: this is not safe, what if there will be `HTTP/2.0`? I would rather do `.split("\\s", 2)[1]`

src/main/java/org/takes/facets/fallback/FbStatus.java
```diff
@@ -45,7 +47,46 @@
...
+                return Collections.<Response>singleton(
+                    new RsFluent(res)
+                        .withHeader("Content-Type", "text/plain")
```
yegor256: it's better to use `RsText`

yegor256:
@komaz a few comments from me above

komaz:
@yegor256 
- The code relies on a length of `"HTTP/1.1 "` string. Since `"HTTP/2.0 "` has the same length, everything will work as expected. But I agree your solution is better, I've pushed an updated version.
- Can't use `RsText` until #310 is fixed, changed to `RsWithType` and improved a test to check response status code. Or should I wait for #310 fix?

yegor256:
@rultor try to merge

davvd:

davvd:
@rultor deploy now pls

elenavolokhova:
@ggajos 
>yegor256 added a note 2 days ago
this is not safe, what if there will be HTTP/2.0? I would rather do .split("\\s", 2)[1]

And couple more of his comments have no addressee in the beginning. In the next ticket please ask him to put username to all comments even if they are left on 'Files changed' tab.

I'm asking you as a performer of this ticket who is responsible for it's quality.

Please confirm that you will try to fix such issues in your future tickets.

elenavolokhova:
@ggajos I need your confirmation in order to complete QA review

ggajos:
@elenavolokhova I can't see this message anywhere here `this is not safe, what if there will be HTTP/2.0? I would rather do .split("\s", 2)[1]`

elenavolokhova:
@ggajos It is in comments left on 'Files Changed' tab. They are visible here in expandable sections like: Show outdated diff yegor256 commented on an outdated diff 7 days ago

It is not obvious. I know. But it is very important for PR creator to receive notification from that page too.

ggajos:
@elenavolokhova ok, thanks, got it now

ggajos:
@yegor256 please remember to address all comments directly

elenavolokhova:

davvd:

@elenavolokhova thanks a lot, next time everybody should try to make it better

davvd:
@ggajos just added 10 mins to @elenavolokhova (for QA), payment ID is `58205785`... Thank you, I have added **22 mins** to your account in payment/transaction "`AP-27333210SG973874N`", time consumed: 95 hours and 25 mins... extra minutes for review comments (c=7)... added +22 to your rating, now it is equal to [+2231](http://www.netbout.com/b/35094?open=rating)

Issue #296, opened 2015-05-18
RsPrettyXML


yegor256:
Let's create `RsPrettyXML` class, which will re-format response body to make it look "pretty" indented:

```java
System.out.println(
  new RsPrint(
    new RsPrettyXML(
      new RsWithBody(
        "<test><a>foo</a></test>"
      )
    )
  ).printBody()
);
```

I expect this result:

```
<test>
  <a>foo</a>
</test>
```

davvd:
@yegor256 tagged this issue with "bug"

davvd:
@yegor256 since there is no milestone yet I set it to "1.0"

davvd:
@yegor256 thank you for reporting this, I added 15 mins to your acc, transaction `000-ceb43c29`

davvd:
@ikhvostenkov the task is yours, please go ahead

ikhvostenkov:
@yegor256 Issue has been fixed with https://github.com/yegor256/takes/pull/308. If you do not have any comments here, could you please close this.

yegor256:
@ikhvostenkov thanks!

yegor256:
@rultor release, tag is `0.17`

yegor256:
@rultor release, tag is `0.17`

davvd:
@ypshenychka please, review this task for compliance with our [quality rules](http://www.teamed.io/qa.html)

ypshenychka:
@davvd Ticket should have assignee. Please assign @ikhvostenkov to this task.

ypshenychka:
@davvd I see there is a corresponding label on the ticket. Please ignore my previous message.

ypshenychka:
@davvd Quality is good here.

davvd:
> @davvd Quality is good here.

@ypshenychka many thanks

davvd:
@ikhvostenkov just added 10 mins to @ypshenychka (for QA), payment ID is `57788744`. I added **30 mins** to your account, many thanks for working with the project! The completion time here was 57788756.. added +30 to your rating, now it is equal to [+145](http://www.netbout.com/b/35232?open=rating)

PR #308, opened 2015-05-21
 RsPrettyXML created

ikhvostenkov:
This is fix for https://github.com/yegor256/takes/issues/296. Added new ``` RsPrettyXML``` class and test ```RsPrettyXMLTest``` for it.

### DIFF
--- a/src/main/java/org/takes/rs/RsPrettyXML.java
+++ b/src/main/java/org/takes/rs/RsPrettyXML.java
@@ -23,11 +23,10 @@
  */
 package org.takes.rs;
 
+import com.jcabi.aspects.Cacheable;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.List;
-import java.util.concurrent.CopyOnWriteArrayList;
 import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerException;
@@ -55,12 +54,6 @@ public final class RsPrettyXML implements Response {
     private final transient Response origin;
 
     /**
-     * Response with properly transformed body.
-     */
-    private final transient List<Response> transformed =
-        new CopyOnWriteArrayList<Response>();
-
-    /**
      * Ctor.
      * @param res Original response
      */
@@ -83,18 +76,12 @@ public final class RsPrettyXML implements Response {
      * @return Response just made
      * @throws IOException If fails
      */
+    @Cacheable
     private Response make() throws IOException {
-        synchronized (this.transformed) {
-            if (this.transformed.isEmpty()) {
-                this.transformed.add(
-                    new RsWithBody(
-                        this.origin,
-                        RsPrettyXML.transform(this.origin.body())
-                    )
-                );
-            }
-        }
-        return this.transformed.get(0);
+        return new RsWithBody(
+            this.origin,
+            RsPrettyXML.transform(this.origin.body())
+        );
     }
 
     /**
--- a/src/main/java/org/takes/rs/RsPrettyXML.java
+++ b/src/main/java/org/takes/rs/RsPrettyXML.java
@@ -1,0 +1,123 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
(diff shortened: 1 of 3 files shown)


ikhvostenkov:
@davvd PR is ready for review.

davvd:
@ikhvostenkov Thanks, I will find someone to review this PR soon

davvd:
@carlosmiranda this pull request is for you,please review

davvd:
> @davvd PR is ready for review.

@ikhvostenkov will take a look, thanks

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -0,0 +1,123 @@
...
+        return this.make().body();
+    }
+
+    /**
+     * Make a response.
+     * @return Response just made
+     * @throws IOException If fails
+     */
+    private Response make() throws IOException {
+        synchronized (this.transformed) {
```
carlosmiranda: @ikhvostenkov Why are we syncing `CopyOnWriteArrayList` operations? The class is [thread safe](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArrayList.html).

src/test/java/org/takes/rs/RsPrettyXMLTest.java
```diff
@@ -0,0 +1,105 @@
...
+            )
+        );
+    }
+
+    /**
+     * RsPrettyXML can compare objects.
+     * @throws Exception If some problem inside
+     */
+    @Test
+    public void comparesObjects() throws Exception {
```
carlosmiranda: @ikhvostenkov Test name is a misnomer, since we aren't "comparing" objects, we're checking whether they're equal or not (big difference in Java). Let's name it `conformsToEqualsAndHashCode`.

src/test/java/org/takes/rs/RsPrettyXMLTest.java
```diff
@@ -0,0 +1,105 @@
...
+                String.format(
+                    "Content-Length: %d",
+                    baos.toByteArray().length
+                )
+            )
+        );
+    }
+
+    /**
+     * RsPrettyXML can compare objects.
```
carlosmiranda: @ikhvostenkov Should be:

> RsPrettyXML can conform to equals and hash code contract.

carlosmiranda:
@ikhvostenkov Some comments above.

src/main/java/org/takes/rs/RsPrettyXML.java
ikhvostenkov: @carlosmiranda I do not agree here, we have check then act here.
```  if (this.transformed.isEmpty()) {
                this.transformed.add( ```
This is compound operation. Actually here are 2 actions and it should be atomic.

ikhvostenkov:
@carlosmiranda Fixed issues with test and commented on ```synchronized```. Please, check.

src/main/java/org/takes/rs/RsPrettyXML.java
carlosmiranda: @ikhvostenkov I don't see why we can't use [addIfAbsent(E)](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CopyOnWriteArrayList.html#addIfAbsent(E)).

carlosmiranda:
@ikhvostenkov Replied to your comment.

src/main/java/org/takes/rs/RsPrettyXML.java
ikhvostenkov: @carlosmiranda True, but I do not think this is good idea transform body to understand this already exists + equals operation takes some time.

ikhvostenkov:
@carlosmiranda Replied.

src/main/java/org/takes/rs/RsPrettyXML.java
carlosmiranda: @ikhvostenkov Actually, let's remove `transformed` altogether and return `new RsWithBody` directly. The current code is not intuitive and we'd prefer things to be readable.

ikhvostenkov: @carlosmiranda What exactly is not intuitive in this code? If we remove caching this will affect performance not in the positive way. We have to call ```transform``` each time. XML transformation is very expensive operation.

carlosmiranda: @ikhvostenkov You're creating a single object on a `CopyOnWriteArrayList`. It's not only unintuitive, it's also a terrible design. If you insist on the caching (which I understand the rationale of), we're not running out of ready-made options. I suggest that you use [Cacheable](http://aspects.jcabi.com/annotation-cacheable.html) from `jcabi-aspects` instead. Rolling our own caching mechanisms is generally a bad idea.

ikhvostenkov: @carlosmiranda Why in this case this design is bad here and acceptable in ```RsGzip```? I can imagine why ```CopyOnWriteArrayList``` is used, because of not deal with ```NULL``` in case of single object.

carlosmiranda: @ikhvostenkov Because the "intent" of the code is not obvious at all; I'd have to "imagine" the original coder's motivations for writing the code. If `RsGzip` has a bad design it's a bug in `RsGzip`. We have an opportunity to improve things here, let's take it, shall we?

ikhvostenkov: @carlosmiranda Yes, this is possible to redo to ```@Cacheable```, but IMHO there is no needs here.

ikhvostenkov:
@carlosmiranda Redone with ```@Cacheable```. Please, review.

carlosmiranda:
@ikhvostenkov Looks good, thanks!

carlosmiranda:
@rultor let's merge

yegor256:
@ikhvostenkov I'm afraid you should roll back to the caching implementation you had before. We're not allowed to use any external libraries in this project.

ikhvostenkov:
@yegor256 Rolled back to synchronized.

carlosmiranda:
@yegor256 @ikhvostenkov All right, there was the reason. My apologies, gentlemen. :)

carlosmiranda:
@rultor merge please

yegor256:
@rultor try to merge

davvd:

davvd:
@rultor deploy now pls

elenavolokhova:
@davvd Quality is good.

davvd:
> @davvd Quality is good.

@elenavolokhova thanks for the review and a good remark

davvd:
@carlosmiranda just added 10 mins to @elenavolokhova (for QA), payment ID is `57786538`

10 mins added to @yegor256 account (our architect), payment ID is `000-58179fc5`, since quality is good here

**21 mins** was added to your account, many thanks for your contribution! The task took 57786545.

extra minutes for review comments (c=6)

+21 added to your rating, at the moment it is: [+4170](http://www.netbout.com/b/35023?open=rating)

Issue #293, opened 2015-05-15
RsWithCookie must combine cookies


yegor256:
At the moment, if I do this:

```java
new RsWithCookie(
  new RsWithCookie(
    new RqFake(),
    "foo", "test1"
  ),
  "bar", "test2"
);
```

I'm getting this:

```
HTTP/1.1 200 OK 
Set-Cookie: foo=test1
Set-Cookie: bar=test2
```

This is not correct and browsers will only read the first `Set-Cookie`. Cookie `bar` won't be set.

Proper response should look like this:

```
HTTP/1.1 200 OK 
Set-Cookie: foo=test1,bar=test2
```

davvd:
@yegor256 I tagged this as "bug"

davvd:
@yegor256 attached this issue to milestone "1.0" (let me know if this is wrong)

davvd:
@bertram25 the task is for you now, follow [these guidelines](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Don't hesitate to ask any technical questions right here. The budget here is **30 mins**, which is exactly how much time will be paid for, when the task is [completed](http://www.xdsd.org/2014/04/15/definition-of-done.html)

davvd:
@yegor256 many thanks for the bug report, I added 15 mins to your acc, transaction `000-2a911432`

Bertram25:
@davvd I'll commit the PR fixing this issue in one or two hours (It is working locally but still need static analysis fixes.). I may have to need more time for the review and the ticket cleanup, though. Can you give me one more day, in case I need it?

Bertram25:
@davvd PR #314 ready for review. :)

carlosmiranda:
@davvd Need a bit more time

davvd:
> @davvd Need a bit more time

@carlosmiranda OK, take your time

Bertram25:
@yegor256 Hi, :)

Should I wait for QA review here, as the PR was merged and comments are following the requested standard, or should I explicitly request it?

Best regards,

yegor256:
@Bertram25 first, you should make sure the author of the ticket closed it, see http://www.yegor256.com/2014/11/24/principles-of-bug-tracking.html

Bertram25:
@yegor256 Thanks. In that case, the PR fixing this issue is merged and the issue comments are clean. Could you close the issue?

yegor256:
@Bertram25 thanks!

davvd:
[...]

PR #314, opened 2015-05-25
Made RsWithCookie able to combine the given value with previous ones. #293

Bertram25:
Also adapted the corresponding test.

### DIFF
--- a/src/main/java/org/takes/rs/RsWithCookie.java
+++ b/src/main/java/org/takes/rs/RsWithCookie.java
@@ -23,6 +23,8 @@
  */
 package org.takes.rs;
 
+import java.io.IOException;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.EqualsAndHashCode;
 import org.takes.Response;
@@ -73,6 +75,18 @@ public final class RsWithCookie extends RsWrap {
     );
 
     /**
+     * Cookie header name.
+     */
+    private static final CharSequence SET_COOKIE = "Set-Cookie";
+
+    /**
+     * Pattern to get the current cookie value from header.
+     */
+    private static final Pattern COOKIE_PTRN = Pattern.compile(
+        String.format("%s: ([^\\s].*)", RsWithCookie.SET_COOKIE)
+    );
+
+    /**
      * Ctor.
      * @param name Cookie name
      * @param value Value of it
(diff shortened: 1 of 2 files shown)


davvd:
@bertram25 Many thanks, I will find someone to review it before we merge

davvd:
@carlosmiranda this pull request is for you, please review

src/main/java/org/takes/rs/RsWithCookie.java
```diff
@@ -125,6 +145,33 @@ private static String make(final CharSequence name,
...
+        final StringBuilder value = new StringBuilder();
+        try {
+            for (final String header : res.head()) {
+                if (header.contains(RsWithCookie.SET_COOKIE)) {
+                    cookie.append(header);
+                    break;
+                }
+            }
+        } catch (final IOException exception) {
+            return value.toString();
```
carlosmiranda: @Bertram25 Why are we swallowing an Exception here and then performing a return?

src/test/java/org/takes/rs/RsWithCookieTest.java
```diff
@@ -38,22 +38,27 @@
 public final class RsWithCookieTest {
 
     /**
-     * RsWithCookie can add cookies.
+     * RsWithCookie can add cookies (even with several decorations).
```
carlosmiranda: @Bertram25 Let's keep this unit test as it is. Instead, add a second test "addsMultipleCookies" with the behavior we're specifically trying to implement.

src/main/java/org/takes/rs/RsWithCookie.java
```diff
@@ -95,27 +109,33 @@ public RsWithCookie(final Response res, final CharSequence name,
...
      */
-    private static String make(final CharSequence name,
+    private static String make(final String previous, final CharSequence name,
         final CharSequence value, final CharSequence... attrs) {
-        final StringBuilder text = new StringBuilder(
+        final StringBuilder text = new StringBuilder();
+        if (!previous.isEmpty()) {
+            text.append(String.format("%s,", previous));
+        }
+        text.append(
```
carlosmiranda: @Bertram25 We can pass `String.format` statement below on this line without reaching the line length limit.

src/main/java/org/takes/rs/RsWithCookie.java
```diff
@@ -125,6 +145,33 @@ private static String make(final CharSequence name,
     }
 
     /**
+     * Retrieve potential previous cookie string.
+     * @param res Current Response
+     * @return Current cookie value or empty string.
+     */
+    @SuppressWarnings("PMD.OnlyOneReturn")
+    private static String getPreviousValue(final Response res) {
```
carlosmiranda: @Bertram25 Let's simply call it `previousValue`.

carlosmiranda:
@Bertram25 Some comments above.

Bertram25:
@carlosmiranda 

Thanks for the review. :)

> Let's simply call it previousValue.
We can pass String.format statement below on this line without reaching the line length limit.
Let's keep this unit test as it is. Instead, add a second test "addsMultipleCookies" with the behavior we're specifically trying to implement.

Those points should now be fixed:

> Why are we swallowing an Exception here and then performing a return?

As for this point, I've been thinking about it, and I think you're right it is not the best thing to do.
Here is the situation:
res.head() can throw an IOException. And if I make previousValue() throw an Exception as well, then I'll end up being forced to make the RsWithCookie constuctor capable of throwing an exception.
As I didn't see that being done on any constructors in takes, I think this makes it pretty obvious this isn't the right thing to do.
Thus, if I want to make it use the exception correctly, I guess I'd have to create a new Response tye within the RsWithCookie class overriding the head() function and replacing the cookie header, with the cost that I will no more be able to use the Without/WithHeader decorations in that constructor.

If that's what desired, well, I'm back to square one but I can do it, the feature can also be seen in a more general point of view and maybe it is time to create a RsReplaceHeader class?

Just tell me. :)

Best regards,

carlosmiranda:
@Bertram25 your suggested strategy in overriding `Response` methods is similar to how other classes (e.g. `RsWithHeaders`) does it. Let's do that.

Bertram25:
@carlosmiranda Thanks a lot for the hints given!
The new commit should fix the last remaining issue.

Bertram25:
@davvd Please give me one more day to give time to the review. Thanks!

src/main/java/org/takes/rs/RsWithCookie.java
```diff
@@ -94,30 +114,70 @@ public RsWithCookie(final CharSequence name, final CharSequence value,
...
-                RsWithCookie.make(
-                    RsWithCookie.checkName(name),
-                    RsWithCookie.checkValue(value),
-                    attrs
-            )
-        )
+            new Response() {
+                @Override
+                public Iterable<String> head() throws IOException {
+                    return RsWithCookie.extend(
```
carlosmiranda: @Bertram25 I think we can move contents of `extend` method in here now.

carlosmiranda:
@Bertram25 Just one more comment. :)

Bertram25:
@carlosmiranda It should be fixed now.
I had to disable a PMD rule complaining about the lack of a super() call, even though the call is actually done.

carlosmiranda:
@Bertram25 Strange error, but thanks, all looking good now.

carlosmiranda:
@rultor Let's merge please

yegor256:
@rultor try to merge

davvd:
@elenavolokhova review this ticket please, for compliance with our [quality rules](http://www.teamed.io/qa.html)

davvd:
@rultor deploy pls

elenavolokhova:
@carlosmiranda 
 
 There is poor description provided and it is not linked to original ticket. Pull request must briefly explain what was done in the branch, so that we can use this information later, for traceability. 
As far as you are the performer of this ticket you are responsible for its quality. Please fix this issue and make sure that everyone follows this rule in your tickets in the future.

carlosmiranda:
@Bertram25 please elaborate the PR description and link to the original ticket.

Bertram25:
@carlosmiranda My apologies for not doing this earlier. The mistake is duly noted. Is it better now?

elenavolokhova:
@Bertram25 @carlosmiranda Thank you!

elenavolokhova:

davvd:

@elenavolokhova thanks for the QA review, we'll work better next time

davvd:
@carlosmiranda just added 10 mins to @elenavolokhova (for QA), payment ID is `58673640`

**21 mins** sent to your balance (ID `58673643`), many thanks! It took 176 hours and 12 mins.

review comments (c=6) added as a bonus

+21 added to your rating, at the moment it is: [+3986](http://www.netbout.com/b/35023?open=rating)

Issue #221, opened 2015-04-29
Basic Authentication


dmzaytsev:
We have not a Pass implementation for Basic Authentication(RFC-2617). I believe it has to be.
This class should  have  BasicAuth as one of a Ctor parameters.
Something like this
```
interface BasicAuth {
  Identity fetch(String user, String passwd);
}
```

yegor256:
@dmzaytsev totally agree, but I would call it `PsBasic`

davvd:
@dmzaytsev this task will get someone's attention soon

davvd:
@dmzaytsev milestone set to `1.0` (correct me if I am wrong)

davvd:
@dmzaytsev thanks for this report, I added 15 mins to your account, in transaction `56387869`

davvd:
@cyberone it's yours,please go ahead

cyberone:
@davvd please assign someone else. Too busy.

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone 30 points deducted from your rating

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone OK, I will try to assign someone else

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone I deducted 30 points from your rating

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone someone else will help in this task, no problem at all

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone 30 points deducted from your rating

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone I will assign somebody else to this issue

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone I deducted 30 points from your rating

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone got it, someone else will be assigned soon

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone -30 points to your rating

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone I will assign somebody else to this issue

cyberone:
@davvd I've got 5 times -30 to my rating, refund 4 of them please.

davvd:
> @davvd please assign someone else. Too busy. 

@cyberone -30 points to your rating

davvd:
[...]

PR #327, opened 2015-06-01
Added an implementation of the Basic Authentication system. Issue #221

antonini:
Added an implementation of the Basic Authentication system. Issue #221

### DIFF
--- a/src/main/java/org/takes/facets/auth/PsBasic.java
+++ b/src/main/java/org/takes/facets/auth/PsBasic.java
@@ -65,29 +65,23 @@ public final class PsBasic implements Pass {
     @Override
     @SuppressWarnings("PMD.AvoidInstantiatingObjectsInLoops")
     public Opt<Identity> enter(final Request request) throws IOException {
-        String authorization = "";
-        String user = "";
-        String pass = "";
-        boolean found = false;
+        BasicAuth auth = new BasicAuth("", "");
         Opt<Identity> identity = new Opt.Empty<Identity>();
         for (final String head : request.head()) {
             if (head.startsWith(AUTH_HEAD)) {
-                authorization = head.split(AUTH_HEAD)[1].trim();
-                authorization = new String(
-                    DatatypeConverter.parseBase64Binary(authorization)
-                );
-                user = authorization.split(":")[0];
-                pass = authorization.substring(user.length() + 1);
-                found = true;
+                auth = this.readAuthContentOnHead(head);
                 break;
             }
         }
-        if (found && this.entry.check(user, pass)) {
+        if (!auth.isEmpty() && this.entry.check(
+            auth.getUser(),
+            auth.getPass()
+        )) {
             final ConcurrentMap<String, String> props =
                 new ConcurrentHashMap<String, String>(0);
             identity = new Opt.Single<Identity>(
                 new Identity.Simple(
-                    String.format("urn:basic:%s", user),
+                    String.format("urn:basic:%s", auth.getUser()),
                     props
                 )
             );
@@ -99,6 +93,24 @@ public final class PsBasic implements Pass {
     public Response exit(final Response response, final Identity identity)
         throws IOException {
         return response;
+    }
+
+    /**
+     * Read authentication content that is received on the head.
+     * @param head Head
+     * @return BasicAuth instance.
+     */
+    private BasicAuth readAuthContentOnHead(final String head) {
+        final String authorization = new String(
+            DatatypeConverter.parseBase64Binary(
+                head.split(AUTH_HEAD)[1].trim()
+            )
+        );
+        final String user = authorization.split(":")[0];
+        return new BasicAuth(
+            user,
+            authorization.substring(user.length() + 1)
+        );
     }
 
     /**
@@ -118,5 +130,60 @@ public final class PsBasic implements Pass {
          * @return If valid it return <code>true</code>.
          */
         boolean check(String user, String pwd);
+    }
+
+    /**
+     * Used to transfer authentication information.
+     *
+     * @author Endrigo Antonini (teamed@endrigo.com.br)
+     * @version $Id$
+     * @since 0.20
+     */
+    private final class BasicAuth {
+
+        /**
+         * User.
+         */
+        private final String user;
+
+        /**
+         * Password.
+         */
+        private final String pass;
+
+        /**
+         * Ctor.
+         * @param username User
+         * @param password Password
+         */
+        public BasicAuth(final String username, final String password) {
+            super();
+            this.user = username;
+            this.pass = password;
+        }
+
+        /**
+         * Return user.
+         * @return User.
+         */
+        public String getUser() {
+            return this.user;
+        }
+
+        /**
+         * Return Password.
+         * @return Password.
+         */
+        public String getPass() {
+            return this.pass;
+        }
+
+        /**
+         * Check if the object is empty.
+         * @return Return <code>true</code> if user and password is empty.
+         */
+        public boolean isEmpty() {
+            return this.getUser().isEmpty() && this.getPass().isEmpty();
+        }
     }
 }
--- a/src/main/java/org/takes/facets/auth/PsBasic.java
+++ b/src/main/java/org/takes/facets/auth/PsBasic.java
@@ -131,48 +131,43 @@ public final class PsBasic implements Pass {
     }
 
     /**
-     * Entry with fixed credentials.
+     * Fake implementation of {@link PsBasic.Entry}.
+     *
+     * <p>The class is immutable and thread-safe.
      *
      * @author Endrigo Antonini (teamed@endrigo.com.br)
      * @version $Id$
      * @since 0.20
      *
      */
-    public static final class Fixed implements PsBasic.Entry {
+    public static final class Fake implements PsBasic.Entry {
 
         /**
-         * Username.
+         * Should we authenticate a user?
          */
-        private final transient String username;
-
-        /**
-         * Password.
-         */
-        private final transient String password;
+        private final transient boolean condition;
 
         /**
          * Ctor.
-         * @param user Username.
-         * @param pwd Password.
+         * @param cond Condition
          */
-        public Fixed(final String user, final String pwd) {
-            this.username = user;
-            this.password = pwd;
+        public Fake(final boolean cond) {
+            this.condition = cond;
         }
 
         @Override
-        public Opt<Identity> enter(final String user, final String pwd) {
-            final Opt<Identity> identity;
-            if (this.username.equals(user) && this.password.equals(pwd)) {
-                identity = new Opt.Single<Identity>(
+        public Opt<Identity> enter(final String usr, final String pwd) {
+            final Opt<Identity> user;
+            if (this.condition) {
+                user = new Opt.Single<Identity>(
                     new Identity.Simple(
-                        String.format("urn:basic:%s", user)
+                        String.format("urn:basic:%s", usr)
                     )
                 );
             } else {
-                identity = new Opt.Empty<Identity>();
+                user = new Opt.Empty<Identity>();
             }
-            return identity;
+            return user;
         }
     }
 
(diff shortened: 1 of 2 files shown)


davvd:
@antonini Let me find a reviewer for this pull request, thanks for submitting it

davvd:
@caarlos0 please review,thanks

antonini:
@davvd and @caarlos0, I saw that I used an API of Java 8. Going to fix that.

caarlos0:
@antonini ok, let me know when you do.

antonini:
@caarlos0 Done. Can you review it?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,122 @@
...
+    public Opt<Identity> enter(final Request request) throws IOException {
+        String authorization = "";
+        String user = "";
+        String pass = "";
+        boolean found = false;
+        Opt<Identity> identity = new Opt.Empty<Identity>();
+        for (final String head : request.head()) {
+            if (head.startsWith(AUTH_HEAD)) {
+                authorization = head.split(AUTH_HEAD)[1].trim();
+                authorization = new String(
```
caarlos0: @antonini assigning multiple times seems like a bad practice to me...

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,122 @@
...
+                    DatatypeConverter.parseBase64Binary(authorization)
+                );
+                user = authorization.split(":")[0];
+                pass = authorization.substring(user.length() + 1);
```
caarlos0: @antonini I would put this entire if content in another method...

src/test/java/org/takes/facets/auth/PsBasicTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.takes.rq.RqFake;
+import org.takes.rq.RqWithHeaders;
+
+/**
+ * Test of {@link PsBasic}.
+ * @author Endrigo Antonini (teamed@endrigo.com.br)
+ * @version $Id$
+ * @since 0.20
+ */
+public class PsBasicTest {
```
caarlos0: @antonini I think there is a missing test for multiple headers and another for no valid headers at all...

caarlos0:
@antonini couple of comments above

src/main/java/org/takes/facets/auth/PsBasic.java
antonini: @caarlos0 Fixed.

src/main/java/org/takes/facets/auth/PsBasic.java
antonini: @caarlos0 Done. It was necessary to create an internal class to store username and password.

src/test/java/org/takes/facets/auth/PsBasicTest.java
antonini: @caarlos0 Done.

antonini:
@caarlos0 Done. Waiting for your review.

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,189 @@
...
+        public String getPass() {
+            return this.pass;
+        }
+
+        /**
+         * Check if the object is empty.
+         * @return Return <code>true</code> if user and password is empty.
+         */
+        public boolean isEmpty() {
+            return this.getUser().isEmpty() && this.getPass().isEmpty();
```
caarlos0: @antonini why would we ever create an auth with empty user and empty password?

antonini: @caarlos0 You are right! Better using Opt than creating another logic to do the same stuff.

antonini:
@caarlos0 Changes are available to be reviewed.

caarlos0:
@rultor merge

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    public interface Entry {
+
+        /**
+         * Check if is a valid user.
+         * @param user User
+         * @param pwd Password
+         * @return If valid it return <code>true</code>.
+         */
+        boolean check(String user, String pwd);
```
yegor256: @antonini looks like it's better to return `Opt<Identity>` here and let our users decide what is the identity, with all its attributes, etc.

yegor256:
@antonini one comment above. besides that, the entire design is very procedural. I would recommend you to read these articles: http://www.yegor256.com/2014/09/16/getters-and-setters-are-evil.html http://www.yegor256.com/2014/11/20/seven-virtues-of-good-object.html

antonini:
@yegor256 Changes available.

yegor256:
@antonini you're using `Opt` too much, it's a wrong idea. In general, this `Opt` class is NOT a replacement of NULL, but you're using it exactly like one :) Try to do it differently, without Opt

antonini:
@yegor256 Source available. Can you review it?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,175 @@
...
+                identity = new Opt.Single<Identity>(
+                    new Identity.Simple(
+                        String.format("urn:basic:%s", user)
+                    )
+                );
+            }
+        } catch (final IOException ex) {
+            identity = new Opt.Empty<Identity>();
+        }
+        return identity;
```
yegor256: @antonini that's too complex and you're losing the original identity returned by the `entry`. just get what `entry` returns and return it from `enter()`. simple as that.

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Entry with fixed credentials.
+     *
+     * @author Endrigo Antonini (teamed@endrigo.com.br)
+     * @version $Id$
+     * @since 0.20
+     *
+     */
+    public static final class Fixed implements PsBasic.Entry {
```
yegor256: @antonini doesn't look like a useful class for other users.. I can't imagine a situation where this class will be necessary in production code. can you?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,175 @@
...
+     * Empty check.
+     *
+     * @author Endrigo Antonini (teamed@endrigo.com.br)
+     * @version $Id$
+     * @since 0.20
+     */
+    public static final class Empty implements PsBasic.Entry {
+
+        @Override
+        public Opt<Identity> check(final String user, final String pwd) {
```
yegor256: @antonini I would call it `enter` instead, since it's not just checking, but "entering"

yegor256:
@antonini a few comments above. but the big problem is that you're implementing only 50% of the Basic Auth protocol. when no `Authenticate` header is present, you should return a 401 response with `WWW-Authenticate` header inside, see https://en.wikipedia.org/wiki/Basic_access_authentication

src/main/java/org/takes/facets/auth/PsBasic.java
antonini: @yegor256 It's a fake authentication with fixed username and password such as `RqFake`. I thought to use `$$Fake` but I didn't found an acronym to `Basic` with two letters. Should I just use `Fake`? Or should I change it to an outer class with a `$$Fake`?

antonini:
@yegor256 I asked about the class above. Can you answer me that before I commit the source?

yegor256:
@antonini ok, let's keep this class

src/main/java/org/takes/facets/auth/PsBasic.java
antonini: @yegor256 It's not possible to simple return it right now because I have to return the `HTTP 401 status` and `WWW-Authenticate` header.

antonini:
@yegor256 Code available to review.

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,201 @@
...
+                        new RqHeaders.Base(request)
+                    ).single("authorization").split(AUTH_HEAD)[1]
+                )
+            ).trim();
+            final String user = decoded.split(":")[0];
+            if (this.entry.enter(
+                user,
+                decoded.substring(user.length() + 1)
+            ).has()) {
+                identity = new Opt.Single<Identity>(
```
yegor256: @antonini I still don't understand, why can't you just use the result of `this.entry.enter()`? why do you need to create a new identity?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,201 @@
...
+                    new Identity.Simple(
+                        String.format("urn:basic:%s", user)
+                    )
+                );
+            }
+        } catch (final IOException ex) {
```
yegor256: @antonini what is this for? why can't you just let the exception bubble up?

antonini:
@yegor256 Removed code complexity.

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,192 @@
...
+            throw new RsForward(
+                new RsWithHeaders(
+                    new RsFlash("access denied", Level.WARNING),
+                    String.format(
+                        "WWW-Authenticate: Basic ream=\"%s\"",
+                        this.realm
+                    )
+                ),
+                HttpURLConnection.HTTP_UNAUTHORIZED,
+                "/login/start"
```
yegor256: @antonini where did you get this? :) we should use `new RqHref.Base(request).href()` here

yegor256:
@antonini one more problem above

antonini:
@yegor256 I changed the URL. can you check it?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -0,0 +1,193 @@
...
+            )
+        ).trim();
+        final String user = decoded.split(":")[0];
+        final Opt<Identity> identity = this.entry.enter(
+            user,
+            decoded.substring(user.length() + 1)
+        );
+        if (!identity.has()) {
+            throw new RsForward(
+                new RsWithHeaders(
```
yegor256: @antonini I would use `RsWithHeader` here and get rid of `String.format()`

src/main/java/org/takes/facets/auth/PsBasic.java
yegor256: @antonini if you name this class `Fixed`, it has to encapsulate an `Identity` and always return it back

yegor256: @antonini or `Opt<Identity>`. My point is that there shouldn't be any `if`, since the name is `Fixed`

yegor256:
@antonini a few more

antonini:
@yegor256 Changed to use `RsWithHeader`.

antonini:
@yegor256 I had not seen the other reviews on the Fixed class. Let me work on that first.

src/main/java/org/takes/facets/auth/PsBasic.java
antonini: @yegor256 Agree. I'll rename it to  `Fake` and use same logic of `PsFake`.

antonini:
@yegor256 Improved the source. Can you check it?

yegor256:
@rultor try to merge

davvd:
@ypshenychka review this ticket please, for compliance with our [quality rules](http://www.teamed.io/qa.html)

davvd:
@rultor please deploy

ypshenychka:
@davvd Quality is good.

davvd:
> @davvd Quality is good.

@ypshenychka thanks for the review and a good remark

davvd:
@caarlos0 10 mins was added to the account of @ypshenychka (for QA review), in transaction 59122289

**18 mins** added to your account (payment number `AP-9V2383495L219121G`), many thanks for your contribution! 287 hours and 44 mins spent here.

you're getting extra minutes here (c=3)

+18 added to your rating, at the moment it is: [+2328](http://www.netbout.com/b/35095?open=rating)

Issue #331, opened 2015-06-04
RqMultipart.Base incorrect handles non-printable ASCII characters


dmzaytsev:
RqMultipart.Base incorrect handles non-printable ASCII characters
to reproduce it just make a small changes in `RqMultipartTest.handlesRequestInTime()`
https://github.com/yegor256/takes/blob/master/src/test/java/org/takes/rq/RqMultipartTest.java#L357

change 
```
for (int ind = 0; ind < length; ++ind) {
            bwr.write("X");
}
```
to
```
Random rnd = new Random();
for (int ind = 0; ind < length; ++ind) {
    bwr.write(String.format("%c", rnd.nextInt(128)));         
}
```

`bwr.write(String.format("%c", rnd.nextInt(108)+20));` will work fine.
so problem in the non-printable characters from 0 to 19

davvd:
@dmzaytsev I will ask someone to take care of this task soon

davvd:
@dmzaytsev since the ticket has no milestone I set it to 1.0

davvd:
@dmzaytsev many thanks for the bug report, I added 15 mins to your acc, transaction `58631336`

davvd:
@lautarobock this task is yours, please help (see our key [principles of work](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html)). If you have any technical questions, don't hesitate to ask right here... The budget here is **30 mins**, which is exactly how much time will be paid for, when the task is [completed](http://www.xdsd.org/2014/04/15/definition-of-done.html)

dmzaytsev:
@lautarobock It would better to create a new test for this issue. Also please add a hash calculation for input and output data and compare them.

dmzaytsev:
@lautarobock any progress here ?

yegor256:
@dmzaytsev let me fix it myself, will be faster. assign it to me pls

dmzaytsev:
@davvd assign this ticket to @yegor256 please

yegor256:
@dmzaytsev maybe you can fix it? I have no idea why it's happening. I only managed to reproduce it in a test, just like you suggested:

dmzaytsev:
@yegor256 ok, I'll try, assign it to me pls

yegor256:
@davvd assign @dmzaytsev to this task pls

davvd:
> @davvd assign @dmzaytsev to this task pls

@yegor256 of course! @dmzaytsev it's your ticket now, please go ahead

[...]

PR #335, opened 2015-06-08
#331 RqMultipart.Base fixed bug with the distortion of the contents

dmzaytsev:
For #331 

* fixed a bug with the distortion of the contents of the request if the request contains the boundary characters in some conditions
* added the unit test to check this.


yegor256:
@rultor merge

PR #348, opened 2015-06-18
#339 RsPrettyXML retains html5 doctype

### LINKED ISSUE #339: RsPrettyXML is losing <!DOCTYPE html>
RsPrettyXML must keep <!DOCTYPE html> on result.

<b>given input</b>

```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
   <head>
    ...
   </head>
   <body>
    ...
   </body>
</html>
```
<b>expected output</b>

```
<!DOCTYPE html>

<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
   
   <head> 
     ...
      
   </head>
   
   <body>
     ...
     
   </body>
   
</html>
```

<b>current output</b>

```
<html xmlns="http://www.w3.org/1999/xhtml" lang="en">
   
   <head>
      ...
   </head>
   
   <body>
    ...
   </body>
   
</html>
```

mbarbieri:
Checks if the provided doctype is for HTML 5, if so it preserves the doctype declaration.
To do that I also needed to upgrade to latest SAX version.

Fixes #339

### DIFF
--- a/src/main/java/org/takes/rs/RsPrettyXML.java
+++ b/src/main/java/org/takes/rs/RsPrettyXML.java
@@ -39,7 +39,6 @@ import javax.xml.transform.stream.StreamResult;
 import lombok.EqualsAndHashCode;
 import org.takes.Response;
 import org.w3c.dom.Document;
-import org.w3c.dom.DocumentType;
 import org.xml.sax.InputSource;
 
 /**
@@ -134,37 +133,28 @@ public final class RsPrettyXML implements Response {
      * Checks if the input is an html5 document.
      * @param body The body to be checked.
      * @return True if the input is an html5 document.
-     * @checkstyle MethodNameCheck (4 lines)
+     * @checkstyle MethodNameCheck (3 lines)
      */
-    @SuppressWarnings({ "PMD.AvoidCatchingGenericException",
-        "PMD.OnlyOneReturn" })
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
     private static boolean isHtml5(final InputStream body) {
-        final Document document;
+        Document document = null;
+        boolean valid = true;
         try {
             body.mark(Integer.MAX_VALUE);
-            document = parseBody(body);
+            final DocumentBuilderFactory factory =
+                DocumentBuilderFactory.newInstance();
+            final DocumentBuilder builder = factory.newDocumentBuilder();
+            document = builder.parse(body);
             body.reset();
             // @checkstyle IllegalCatchCheck (1 line)
         } catch (final Exception ex) {
-            return false;
+            valid = false;
         }
-        final DocumentType doctype = document.getDoctype();
-        return doctype != null
-            && "html".equalsIgnoreCase(doctype.getName())
-            && doctype.getSystemId() == null
-            && doctype.getPublicId() == null;
-    }
-
-    /**
-     * Parses the input stream and returns the Document built.
-     * @param body The body to be parsed.
-     * @return The document built.
-     * @throws Exception if something goes wrong.
-     */
-    private static Document parseBody(final InputStream body) throws Exception {
-        final DocumentBuilderFactory factory =
-            DocumentBuilderFactory.newInstance();
-        final DocumentBuilder builder = factory.newDocumentBuilder();
-        return builder.parse(body);
+        // @checkstyle BooleanExpressionComplexityCheck (5 lines)
+        return valid
+            && document.getDoctype() != null
+            && "html".equalsIgnoreCase(document.getDoctype().getName())
+            && document.getDoctype().getSystemId() == null
+            && document.getDoctype().getPublicId() == null;
     }
 }
--- a/src/main/java/org/takes/rs/RsPrettyXML.java
+++ b/src/main/java/org/takes/rs/RsPrettyXML.java
@@ -134,10 +134,10 @@ public final class RsPrettyXML implements Response {
      * Checks if the input is an html5 document.
      * @param body The body to be checked.
      * @return True if the input is an html5 document.
-     * @checkstyle MethodNameCheck (2 lines)
+     * @checkstyle MethodNameCheck (4 lines)
      */
-    @SuppressWarnings({"PMD.AvoidCatchingGenericException",
-        "PMD.OnlyOneReturn"})
+    @SuppressWarnings({ "PMD.AvoidCatchingGenericException",
+        "PMD.OnlyOneReturn" })
     private static boolean isHtml5(final InputStream body) {
         final Document document;
         try {
--- a/src/main/java/org/takes/rs/RsPrettyXML.java
+++ b/src/main/java/org/takes/rs/RsPrettyXML.java
@@ -39,6 +39,7 @@ import javax.xml.transform.sax.SAXSource;
 import javax.xml.transform.stream.StreamResult;
 import lombok.EqualsAndHashCode;
 import org.takes.Response;
+import org.takes.misc.Opt;
 import org.w3c.dom.Document;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
@@ -50,6 +51,7 @@ import org.xml.sax.SAXException;
  * @author Igor Khvostenkov (ikhvostenkov@gmail.com)
  * @version $Id$
  * @since 1.0
+ * @checkstyle ClassDataAbstractionCouplingCheck (100 lines)
  */
 @EqualsAndHashCode(of = "origin")
 public final class RsPrettyXML implements Response {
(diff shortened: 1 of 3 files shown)


davvd:
@mbarbieri Thanks, I will find someone to review this PR soon

davvd:
@carlosmiranda this pull request is for you, please review

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +118,53 @@ private Response make() throws IOException {
...
     }
+
+    /**
+     * Checks if the input is an html5 document.
+     * @param body The body to be checked.
+     * @return True if the input is an html5 document.
+     * @checkstyle MethodNameCheck (4 lines)
+     */
+    @SuppressWarnings({ "PMD.AvoidCatchingGenericException",
+        "PMD.OnlyOneReturn" })
```
carlosmiranda: @mbarbieri Let's have a variable that will store the return value instead of having two different `return` statements.

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +118,53 @@ private Response make() throws IOException {
...
+    private static boolean isHtml5(final InputStream body) {
+        final Document document;
+        try {
+            body.mark(Integer.MAX_VALUE);
+            document = parseBody(body);
+            body.reset();
+            // @checkstyle IllegalCatchCheck (1 line)
+        } catch (final Exception ex) {
+            return false;
```
carlosmiranda: @mbarbieri Let's log the exception that's being thrown in a `INFO` or `DEBUG` message.

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +118,53 @@ private Response make() throws IOException {
...
         return result.toByteArray();
     }
+
+    /**
+     * Checks if the input is an html5 document.
+     * @param body The body to be checked.
+     * @return True if the input is an html5 document.
+     * @checkstyle MethodNameCheck (4 lines)
+     */
+    @SuppressWarnings({ "PMD.AvoidCatchingGenericException",
```
carlosmiranda: @mbarbieri Why not catch a more specific exception type?

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +118,53 @@ private Response make() throws IOException {
...
+            && doctype.getPublicId() == null;
+    }
+
+    /**
+     * Parses the input stream and returns the Document built.
+     * @param body The body to be parsed.
+     * @return The document built.
+     * @throws Exception if something goes wrong.
+     */
+    private static Document parseBody(final InputStream body) throws Exception {
```
carlosmiranda: @mbarbieri We're only using this method once, let's just include its contents inline with `isHtml5` instead of having it separately.

carlosmiranda:
@mbarbieri Some comments above.

src/main/java/org/takes/rs/RsPrettyXML.java
mbarbieri: @carlosmiranda done

src/main/java/org/takes/rs/RsPrettyXML.java
mbarbieri: @carlosmiranda it's not a "real" exception... it's the only way to know if the input is a proper xml file. If SAX had a `parse` function returning a boolean instead of throwing exceptions it would be better

src/main/java/org/takes/rs/RsPrettyXML.java
mbarbieri: @carlosmiranda again because it's not a "real" exception. In both cases I just need to return false.

src/main/java/org/takes/rs/RsPrettyXML.java
mbarbieri: @carlosmiranda ok done. Even though I prefer having a separate method to improve readability

mbarbieri:
@carlosmiranda Some answers above.

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +117,44 @@ private Response make() throws IOException {
...
+        boolean valid = true;
+        try {
+            body.mark(Integer.MAX_VALUE);
+            final DocumentBuilderFactory factory =
+                DocumentBuilderFactory.newInstance();
+            final DocumentBuilder builder = factory.newDocumentBuilder();
+            document = builder.parse(body);
+            body.reset();
+            // @checkstyle IllegalCatchCheck (1 line)
+        } catch (final Exception ex) {
```
carlosmiranda: @mbarbieri What do you mean it's not a "real exception"? Why not catch `SAXException` and `IOException` separately? Seems to me that they are thrown in different situations - see [javadoc](http://docs.oracle.com/javase/7/docs/api/javax/xml/parsers/DocumentBuilder.html#parse(java.io.InputStream)).

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +117,44 @@ private Response make() throws IOException {
...
+     * @return True if the input is an html5 document.
+     * @checkstyle MethodNameCheck (3 lines)
+     */
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    private static boolean isHtml5(final InputStream body) {
+        Document document = null;
+        boolean valid = true;
+        try {
+            body.mark(Integer.MAX_VALUE);
+            final DocumentBuilderFactory factory =
```
carlosmiranda: @mbarbieri We're not using `factory` anywhere else, let's call `newDocumentBuilder()` directly after `newInstance()`.

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +117,44 @@ private Response make() throws IOException {
...
+                DocumentBuilderFactory.newInstance();
+            final DocumentBuilder builder = factory.newDocumentBuilder();
```
carlosmiranda: @mbarbieri Same here, the `builder` is not used anywhere else, we can call `parse` directly and assign it to `document`.

carlosmiranda:
@mbarbieri Some more comments from me

mbarbieri:
@carlosmiranda done

carlosmiranda:
@rultor merge

src/main/java/org/takes/rs/RsPrettyXML.java
```diff
@@ -114,10 +119,46 @@ private Response make() throws IOException {
...
+    /**
+     * Checks if the input is an html5 document.
+     * @param body The body to be checked.
+     * @return True if the input is an html5 document.
+     * @checkstyle MethodNameCheck (4 lines)
+     * @throws IOException If there are problems reading the input.
+     */
+    @LogExceptions
+    private static boolean isHtml5(final InputStream body) throws IOException {
+        Document document = null;
```
yegor256: @mbarbieri we never use NULL, refactor this method somehow

yegor256:
@mbarbieri one comment above

mbarbieri:
@yegor256 null removed

yegor256:
@mbarbieri I don't think this is a good idea... what if the document is not HTML5? we will lose it's doctype? there are many other options: https://en.wikipedia.org/wiki/Document_type_declaration I think we should think about something more generic and correct.

mbarbieri:
@yegor256 right... what about creating a:
```java
class Document {
   public Document(Transformer t) {...}
   public Result transformUsing(Transformer t) {
    // check which type of document I am (html5 or something else)
   // and perform the transformation
  }
}
```
and set some parameter to the transformer (like indent) before passing it to the Document object?

yegor256:
@mbarbieri I'm not sure I understood you... can you explain better?

mbarbieri:
@yegor256 I mean that since SAX doesn't have any way (AFAIK) to simply check the doctype, the only way seems to be to try to parse it and see if it throws some exception.
So I was thinking to wrap the document in an object that can check its type by trying :disappointed:  and than transform itself with the proper doctype configured

yegor256:
@mbarbieri I still don't understand. if you think you can do it - go ahead and create a pull request. but be ready to see it rejected. if you don't know how to do it, just reject the task.

carlosmiranda:
@mbarbieri Any updates here?

carlosmiranda:
@mbarbieri I'd like to know if you have updates for us.

carlosmiranda:
@mbarbieri If you're not continuing with this ticket please close the PR.

carlosmiranda:
@mbarbieri @yegor256 can we close this now? I believe this is not going to be continued anymore...

yegor256:
@carlosmiranda I'm closing it

davvd:
@ypshenychka please, let us know what do you think about this ticket, according to our [QA rules](http://www.teamed.io/qa.html)

ypshenychka:
@davvd Quality is good here.

davvd:

@ypshenychka thank you

davvd:
@carlosmiranda I added 10 mins to @ypshenychka (for QA review) in transaction 64922685

paid, thanks, added **22 mins** to your account, payment `64922695`, 1990 hours and 29 mins was spent

you're getting extra minutes here (c=7)

+22 to your rating, your total score is [+3874](http://www.netbout.com/b/35023?open=rating)

PR #365, opened 2015-07-20
Implement PsBasic.Default

### LINKED ISSUE #349: PsBasic misses Default Entry implementation
Recently it was introduced Basic Authentication `Pass` implementations with #221 called `PsBasic`. As part of implementation `PsBasic.Entry` interface was introduced with `PsBasic.Fake` and `PsBasic.Empty` implementations.

I think `PsBasic` misses the `Default` implementation in order to be able to use this `Pass` implementation having real username/password pairs.

Suseika:
Implement `PsBasic.Default`. This class is a Pass that maintains a collection of login/password pairs and checks a login attempt for correctness.

Fixes #349


Suseika:
@davvd Please find a reveiwer.

davvd:
@suseika Thanks for your pull request, let me find someone who can review it

davvd:
@ggajos review this one, please

src/test/java/org/takes/facets/auth/PsBasicDefaultTest.java
```diff
@@ -0,0 +1,122 @@
...
+                new Identity.Simple("urn:foo:alice")
+            )
+        );
+    }
+    /**
+     * {@link org.takes.facets.auth.PsBasic.Default} accepts correct
+     * login/password pair.
+     */
+    @Test
+    public void defaultPassCanAcceptCorrectLoginPasswordPair() {
```
ggajos: @Suseika please see http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#test-method-names

src/test/java/org/takes/facets/auth/PsBasicDefaultTest.java
```diff
@@ -0,0 +1,122 @@
...
+            Matchers.equalTo(URN)
+        );
+    }
+
+    /**
+     * {@link org.takes.facets.auth.PsBasic.Default} rejects incorrect
+     * passowrd even with an existing login.
+     */
+    @Test
+    public void defaultPassCanRejectIncorrectPassword() {
```
ggajos: @Suseika please see http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#test-method-names

src/test/java/org/takes/facets/auth/PsBasicDefaultTest.java
```diff
@@ -0,0 +1,122 @@
...
+            Matchers.is(false)
+        );
+    }
+
+    /**
+     * {@link org.takes.facets.auth.PsBasic.Default} rejects non-existing
+     * login.
+     */
+    @Test
+    public void defaultPassCanRejectIncorrectLogin() {
```
ggajos: @Suseika please see http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#test-method-names

ggajos:
@Suseika few comments

Suseika:
@ggajos Fixed

ggajos:
@rultor merge please

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -185,4 +187,44 @@ public Fake(final boolean cond) {
...
+        @Override
+        public Opt<Identity> enter(final String user, final String pwd) {
+            final Opt<Identity> identity;
+            final User valid = this.usernames.get(user);
+            if (valid == null) {
+                identity = new Opt.Empty<Identity>();
+            } else if (valid.password().equals(pwd)) {
+                identity = new Opt.Single<Identity>(valid.identity());
+            } else {
+                identity = new Opt.Empty<Identity>();
```
yegor256: @Suseika hm.. looks like code duplication to me. how about this:

```
if (valid == null || !valid.password().equals(pwd)) {
} else {
}
```

yegor256:
@Suseika I don't like this design. Class `User` belongs only to `PsBasic`, but you made it public in the package. Why would anyone else need it? Besides, I would NOT use such a class, but instead find a way to manage these login/passwords inside `PsBasic.Default`. how about this:

```
new PsBasic.Default(
  "yegor password urn:test:123",
  "peter super%20secret%20password urn:foo:peter"
)
```

We don't need to have full identities here, with properties. It's a simple class for basic needs. If someone needs something bigger, he will implement it from scratch

Suseika:
@yegor256 But that wouldn't be a proper object-oriented solution. Why would we hold multiple data items in a single string when we are writing in an object oriented language? That's similar to byte-hacking when integer constants are used instead of enums, and instead of deriving one enum instance from another we'd do some obscure bitwise operations, you know what I mean. If we pack everything in a string, we will need additional logic to validate that string, i.e. that it consists of 3 pieces. On the other hand, `User` class makes it explicit what you have to provide to create a `PsBasic.Default` instance. Even with your example you demonstrated that there is an issue with spaces (or whatever character is chosen as a separator).

If we don't need full Identities with properties, how about changing User constructor parameters to `User(String login, String password, String urn)` with the third parameter being a String instead of Identity?

> Class User belongs only to `PsBasic`, but you made it public in the package. Why would anyone else need it? 

To construct instances of `PsBasic.Default`. I explained above why I think that working with objects is better than working with strings.

I think that your solution would create much more problems than it would solve. So I'll ask before beginning rewriting the class: are you sure you want me to reimplement `PsBasic.Default` to have Strings instead of Users in its constructor?

yegor256:
@Suseika you're right about OOP approach, but at the same time wrong. Here is why. This `User` class is not a class really, but just a holder of data. A bag where we carry some other objects. Right? And this is what objects should NOT be used for. This class has no behavior. It just holds three strings. That's totally against OOP.

On the other hand, the text-based approach I suggested is purely data-focused and compact.

If you want to be truly object oriented, this is how your design should look:

```
new PsBasic.Default(
  new Users(
    new Users.Entry(
      "jeffrey", new User("super secret", new Identity.Simple("..."))
    ).
    new Users.Entry(
      "peter", new User("my password", new Identity.Simple("..."))
    )
  )
)
```

Then, here is how `User` should look:

```java
interface User {
  Opt<Identity> authenticate(String password);
}
```

This is true OOP, but it's so verbose... That's why I suggested strings.

Suseika:
@yegor256 I reimplemented it how you suggested, please take a look.

pom.xml
```diff
@@ -221,7 +221,6 @@
             <groupId>commons-lang</groupId>
             <artifactId>commons-lang</artifactId>
             <version>2.4</version>
-            <scope>test</scope>
```
yegor256: @Suseika we can't use **any** dependencies here

yegor256:
@Suseika we should not use any dependencies in this project, in `compile` scope

Suseika:
@yegor256 Done, see the new commit.

src/test/java/org/takes/facets/auth/PsBasicDefaultTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * Password of one of users.
+     */
+    private static final String BOB_PASSWORD = "qwerty";
+
+    /**
+     * Existing users.
+     */
+    private final transient String[] users = new String[] {
```
yegor256: @Suseika that's a bad idea in general, to couple tests through the shared data, see this: http://www.yegor256.com/2015/05/25/unit-test-scaffolding.html

yegor256:
@Suseika looks good, but test design is wrong, see my comment. Also, try more complex input data. For example, try password `"hey \u20ac"`

Suseika:
@yegor256 `"hey \u20ac"` is not a valid Java string. If you mean password `"hey \u0020ac", then `PsBasic.Default` constructor will fail, because we decided to choose space as a delimiter, and that string is just "hey  ac" string with two spaces instead of one, so it will fail if it is used in `"username hey \u0020ac urn:foo:user"`. If you insist on the text-based approach, do you have any suggestions on how to treat spaces in passwords?

> Also, try more complex input data.

I made passwords a bit more varying (used punctuation and other languages), is that enough?

Please see the new commit.

yegor256:
@Suseika what if the user wants to use a password with a space inside?

Suseika:
@yegor256 I know, that is exactly my concern. You suggested the text format "username password urn" with spaces as separators, but it clearly is not appropriate for the task. And any character could potentially be used in a password, so that would always be a problem whatever we choose as a separator character. But you don't like the OO approach because it is either too verbose or objects turn out to be data bags. So how do you want me to implement passing users into `PsBasic.Default` constructor?

Suseika:
@yegor256 I personally would go with the "objects as data bags" approach I was trying to use earlier because it clearly represents what is expected as input to PsBasic.Default and doesn't suffer from the separator character problem. I understand why you don't like it, but it is the best alternative I can come up with.

yegor256:
@Suseika we just have to URL-encode each of three parts and that's it.

Suseika:
@yegor256 I implemented what you suggested. Please see the commits.

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -185,4 +189,158 @@ public Fake(final boolean cond) {
...
+     * @since 0.22
+     */
+    public static final class Default implements PsBasic.Entry {
+
+        /**
+         * How keys in
+         * {@link org.takes.facets.auth.PsBasic.Default#usernames} are
+         * formatted.
+         */
+        public static final String KEY_FORMAT = "%s %s";
```
yegor256: @Suseika why it's public?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -185,4 +189,158 @@ public Fake(final boolean cond) {
...
+
+        /**
+         * Encoding for URLEncode#encode.
+         */
+        public static final String ENCODING = "UTF-8";
```
yegor256: @Suseika same here, why public? see http://www.yegor256.com/2015/07/06/public-static-literals.html

yegor256:
@Suseika looks perfect now, one small comment above

Suseika:
@yegor256 Sorry, I simply overlooked this. Of course it has to be private. By the way, shouldn't qulice disallow `public static` fields? Or are there instances when those can be tolerable? Can I suggest it as a feature to qulice repo?

Fixed, please see above commit.

yegor256:
@Suseika I think it will be too harsh for qulice :)

yegor256:
@rultor try to merge

davvd:
@ypshenychka please, let us know what do you think about this ticket, according to our [QA rules](http://www.teamed.io/qa.html)

davvd:
@rultor please deploy

ypshenychka:
@ggajos 
According to our [QA Rules](http://www.teamed.io/qa.html):

Issues found are moslty cosmetic. Try to find at least three major problems while future reviews.

Please confirm that everything is clear and you have no questions regarding this approach.

ggajos:
@ypshenychka I will, thanks.

ypshenychka:
@ggajos Thank you!

ypshenychka:
@davvd Quality is acceptable.

davvd:
> @davvd Quality is acceptable.

@ypshenychka thank you, let's try to make it "good" next time

davvd:
@ggajos I added 10 mins to @ypshenychka (for QA review) in transaction 62124492. Thanks for your contribution, **18 mins** was added to your account, payment ID is `AP-7N617714DT396041L`, 175 hours and 5 mins spent total. review comments (c=3) added as a bonus. +18 to your rating, your total score is [+588](http://www.netbout.com/b/35094?open=rating)

Issue #306, opened 2015-05-20
persistent connections


yegor256:
At the moment we don't support [HTTP persistent connections](https://en.wikipedia.org/wiki/HTTP_persistent_connection). Would be great to implement this feature.

davvd:
@yegor256 added "bug" tag to this issue

davvd:
@yegor256 milestone set to `1.0` (correct me if I am wrong)

davvd:
@yegor256 thanks for reporting! I topped your account for 15 mins, transaction 000-cfcfbe88

davvd:
@adko-pl please go ahead, it's your task now, keep [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) in mind, and don't hesitate to ask any technical questions you may have... The cost of this task is **30 mins** (this is exactly how much will be paid, not less not more), when the task is [done](http://www.xdsd.org/2014/04/15/definition-of-done.html)

adko-pl:
@davvd please find another developer

davvd:
> @davvd please find another developer

@adko-pl -30 points to your rating :(

davvd:
> @davvd please find another developer

@adko-pl got it, someone else will be assigned soon

davvd:
@lautarobock can you please help? Keep in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). If you have any technical questions, don't hesitate to ask right here; Total fixed cost of this task is **30 mins** (see [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for more info)

lautarobock:
PR #337 is ready for review

davvd:
> @davvd  PR #337 is ready for review

@lautarobock I will take a look, thanks

lautarobock:
@davvd any news?

davvd:
@lautarobock you're working with this issue for 15 days already

-30 to your rating, your total score is [+90](http://www.netbout.com/b/35344?open=rating)

davvd:
@lautarobock the task is overdue, and I have to re-assign it to someone else. Please, stop working with it immediately. In general, we're against overdue tasks, check [this page](http://www.xdsd.org/2014/04/13/no-obligations-principle.html)

-60 added to your rating, current score is: [-90](http://www.netbout.com/b/35344?open=rating)

davvd:
@erimerturk the task is yours plz proceed

[...]

PR #337, opened 2015-06-08
#306: Modify BkBasic for persistent conections according to HTTP 1.1

lautarobock:
#306 Modify BkBasic checking headers and modifying resposes for http conections according to HTTP 1.1

PR #374, opened 2015-08-06
306 : Support http persistent connections

mstzn36:
#306  :  Support http persistent connections. 
If the client sends a request which has  connection: keep-alive parameter then the server adds connection:keep-alive to the header of the response. 
If the client sends a request which has connection:close parameter then the server don't add connection:keep-alive to the header of the response. 

Puzzle: In this process the server should use the same socket for new coming requests if the connection is persistent.
And the server should create the new socket for new coming requests if the connection is not persistent.

### DIFF
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -84,6 +84,10 @@ public final class BkBasic implements Back {
         this.take = tks;
     }
 
+    // @todo #306 30min/DEV : the server should use same socket for new coming
+    // requests if the connection is persistent.
+    // And the server should create new socket for new coming requests if the
+    // connection is not persistent
     @Override
     public void accept(final Socket socket) throws IOException {
         final InputStream input = socket.getInputStream();
@@ -96,9 +100,7 @@ public final class BkBasic implements Back {
                     keepAlive
             );
         } finally {
-            if (!keepAlive) {
-                input.close();
-            }
+            input.close();
         }
     }
 
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -32,15 +32,19 @@ import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.net.HttpURLConnection;
 import java.net.Socket;
+import java.util.Iterator;
+import java.util.Locale;
 import lombok.EqualsAndHashCode;
 import org.takes.HttpException;
 import org.takes.Request;
 import org.takes.Response;
 import org.takes.Take;
+import org.takes.rq.RqHeaders;
 import org.takes.rq.RqLive;
 import org.takes.rq.RqWithHeaders;
 import org.takes.rs.RsPrint;
 import org.takes.rs.RsText;
+import org.takes.rs.RsWithHeader;
 import org.takes.rs.RsWithStatus;
 
 /**
@@ -58,6 +62,16 @@ import org.takes.rs.RsWithStatus;
 public final class BkBasic implements Back {
 
     /**
+     * Keep alive header.
+     */
+    private static final String CONNECTION = "Connection";
+
+    /**
+     * Keep alive header value.
+     */
+    private static final String KEEP_ALIVE = "Keep-Alive";
+
+    /**
      * Take.
      */
     private final transient Take take;
@@ -73,30 +87,60 @@ public final class BkBasic implements Back {
     @Override
     public void accept(final Socket socket) throws IOException {
         final InputStream input = socket.getInputStream();
+        final RqLive req = new RqLive(input);
+        final boolean keepAlive = this.isConnectionKeepAlive(req);
         try {
             this.print(
-                BkBasic.addSocketHeaders(
-                    new RqLive(input),
-                    socket
-                ),
-                new BufferedOutputStream(socket.getOutputStream())
+                    BkBasic.addSocketHeaders(req, socket),
+                    new BufferedOutputStream(socket.getOutputStream()),
+                    keepAlive
             );
         } finally {
-            input.close();
+            if (!keepAlive) {
+                input.close();
+            }
         }
+    }
+
+    /**
+     * Whether to http connection is persistent or not.
+     * @param req RqLive
+     * @return Boolean
+     * @throws IOException if fails
+     */
+    private boolean isConnectionKeepAlive(final RqLive req) throws IOException {
+        boolean keep = false;
+        final RqHeaders.Base rqHeaders = new RqHeaders.Base(req);
+        if (rqHeaders.head().iterator().hasNext()) {
+            final Iterator<String> values = rqHeaders.header(BkBasic.CONNECTION)
+                    .iterator();
+            if (values.hasNext()) {
+                do {
+                    keep = BkBasic.KEEP_ALIVE.toLowerCase(
+                            Locale.ENGLISH
+                            ).equals(
+                            values.next().toLowerCase(Locale.ENGLISH)
+                    );
+                } while (!keep && values.hasNext());
+            }
+        }
+        return keep;
     }
 
     /**
      * Print response to output stream, safely.
      * @param req Request
-     * @param output Output
-     * @throws IOException If fails
+     * @param output OutputStream
+     * @param keep Boolean
+     * @throws IOException if fails
      */
     @SuppressWarnings("PMD.AvoidCatchingThrowable")
-    private void print(final Request req, final OutputStream output)
-        throws IOException {
+    private void print(final Request req, final OutputStream output,
+                       final boolean keep)
+            throws IOException {
         try {
-            new RsPrint(this.take.act(req)).print(output);
+            final Response res = this.retrieveResponse(req, keep);
+            new RsPrint(res).print(output);
         } catch (final HttpException ex) {
             new RsPrint(BkBasic.failure(ex, ex.code())).print(output);
             // @checkstyle IllegalCatchCheck (7 lines)
@@ -110,6 +154,29 @@ public final class BkBasic implements Back {
         } finally {
             output.close();
         }
+    }
+
+    /**
+     * Create Response by Request.
+     *
+     * @param req Request
+     * @param keep Keep alive
+     * @return Response
+     * @throws IOException if fails
+     */
+    private Response retrieveResponse(final Request req, final boolean keep)
+            throws IOException {
+        final Response res;
+        if (keep) {
+            res = new RsWithHeader(
+                    this.take.act(req),
+                    BkBasic.CONNECTION,
+                    BkBasic.KEEP_ALIVE
+            );
+        } else {
+            res = this.take.act(req);
+        }
+        return res;
     }
 
     /**
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -84,10 +84,15 @@ public final class BkBasic implements Back {
         this.take = tks;
     }
 
-    // @todo #306 15min/DEV : the server should use same socket for new coming
-    // requests if the connection is persistent.
-    // And the server should create new socket for new coming requests if the
-    // connection is not persistent
+    /**
+     * Read from socket.
+     * @param socket TCP socket with HTTP
+     * @throws IOException if some problem occurs
+     * @todo #306 15min/DEV : the server should use same socket for new coming
+     *  requests if the connection is persistent.
+     *  And the server should create new socket for new coming requests if the
+     *  connection is not persistent
+     */
     @Override
     public void accept(final Socket socket) throws IOException {
         final InputStream input = socket.getInputStream();
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -84,7 +84,7 @@ public final class BkBasic implements Back {
         this.take = tks;
     }
 
-    // @todo #306 30min/DEV : the server should use same socket for new coming
+    // @todo #306 15min/DEV : the server should use same socket for new coming
     // requests if the connection is persistent.
     // And the server should create new socket for new coming requests if the
     // connection is not persistent
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -100,9 +100,9 @@ public final class BkBasic implements Back {
         final boolean keepAlive = this.isConnectionKeepAlive(req);
         try {
             this.print(
-                    BkBasic.addSocketHeaders(req, socket),
-                    new BufferedOutputStream(socket.getOutputStream()),
-                    keepAlive
+                BkBasic.addSocketHeaders(req, socket),
+                new BufferedOutputStream(socket.getOutputStream()),
+                keepAlive
             );
         } finally {
             input.close();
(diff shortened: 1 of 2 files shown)


davvd:
@mstzn36 Thanks for your pull request, let me find someone who can review it

davvd:
@dmzaytsev this pull request is for you, please review

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +99,70 @@ public void handlesSocket() throws IOException {
...
     /**
+     * BkBasic can handle HTTP persistent connections.
+     * @throws Exception If some problem inside
+     */
+    @Test
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    public void handlesPersistentConnection() throws Exception {
+        final int port = 8080;
+        final String uri = String.format("http://localhost:%d", port);
+        // @checkstyle MagicNumberCheck (1 line)
```
dmzaytsev: @mstzn36 do we really need this suppression?

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +99,70 @@ public void handlesSocket() throws IOException {
...
+                conn.getHeaderFields(),
+                Matchers.hasKey(connection)
+        );
+        MatcherAssert.assertThat(
+                conn.getHeaderFields().get(connection),
+                Matchers.hasItem(keepAlive)
+        );
+        completed.countDown();
+        completed.await(1L, TimeUnit.MILLISECONDS);
+        MatcherAssert.assertThat(completed.getCount(), Matchers.equalTo(0L));
```
dmzaytsev: @mstzn36 what exactly we check here? 
I think `completed.getCount()` always will be equals 0

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -101,9 +178,9 @@ public void returnsProperResponseCodeOnInvalidUrl() throws IOException {
                 @Override
                 public void exec(final URI home) throws IOException {
                     new JdkRequest(String.format("%s/path/c", home))
-                        .fetch()
-                        .as(RestResponse.class)
-                        .assertStatus(HttpURLConnection.HTTP_NOT_FOUND);
+                            .fetch()
```
dmzaytsev: @mstzn36 why the indentation changed here?

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -73,30 +87,60 @@ public BkBasic(final Take tks) {
...
     @SuppressWarnings("PMD.AvoidCatchingThrowable")
-    private void print(final Request req, final OutputStream output)
-        throws IOException {
+    private void print(final Request req, final OutputStream output,
+                       final boolean keep)
+            throws IOException {
         try {
-            new RsPrint(this.take.act(req)).print(output);
+            final Response res = this.retrieveResponse(req, keep);
+            new RsPrint(res).print(output);
```
dmzaytsev: @mstzn36 `res` can be inline, right?

dmzaytsev:
@mstzn36 please see 4 comments above

dmzaytsev:
@mstzn36 also I believe your test `handlesPersistentConnection() ` not completed.
You need to be sure `FtBasic` reuse the connection. I would suggest use decorator of `BkBasic' in the test. You should make at least two requests to the server and check `BkBasic` uses same socket.

dmzaytsev:
@mstzn36 according our [quality rules](http://www.teamed.io/qa.html)

please add the link to the original ticket to the pull request description

src/test/java/org/takes/http/BkBasicTest.java
mstzn36: @dmzaytsev  i need this suppression because of quality rules. To be compliant to J2EE, a webapp should not use any thread .

dmzaytsev: @mstzn36 I meant MagicNumberCheck

mstzn36:
@dmzaytsev i made some changes according to review. Take a look please.
I think i don't need check that two requests use same socket. And i don't know how i can test 

I just check response header whether it has connection keep-alive parameter. 
If a request has connection keep alive parameter then server doesn't close connection after the current request/response is complete.
If a request has connection closed parameter then the server closes connection after the current request/response is complete.

dmzaytsev:
@mstzn36 I'm not sure if you have added some headers and prevented the socket from closing so it will be enough to implement persistent connections. That's why I asked  to test it. btw you will get a resource leak if you never close the socket. It should be closed by client or by timeout or by some resource limit...
anyway let's ask the architect.
@yegor256 do you think @mstzn36  on right way?
Maybe it would better to use NIO channels/selectors in this case?

yegor256:
@dmzaytsev I don't know... looks difficult to understand :( maybe it's better to create just a test in this ticket and "puzzle" the implementation?

mstzn36:
@dmzaytsev @yegor256 TCP connection pooling and keep-alive are behind the scenes :( So i think it is enough that i just check request and response header in the test. fyi.

dmzaytsev:
@yegor256 sure, I think so

@mstzn36 I didn't mean you have to implement all job in one PR...
Let's follow Yegor's suggestion.
please remove the condition for closing the socket from `BkBasic'(the socket should be closed always), put a puzzle instead. In the unit test make  one connection and check the headers.

dmzaytsev:
@mstzn36 Pull request description [should](http://www.teamed.io/qa.html) explain the solution proposed.
you've created the unit test and the puzzle

mstzn36:
@dmzaytsev please take a look.

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,16 +84,20 @@ public BkBasic(final Take tks) {
         this.take = tks;
     }
 
+    // @todo #306 15min/DEV : the server should use same socket for new coming
```
dmzaytsev: @mstzn36  `@todo #306:30min/DEV`  You missed `:`
please use javadoc-style for the comment
```
/**
*
*/
```

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +102,59 @@ public void handlesSocket() throws IOException {
...
+                            );
+                        } catch (final IOException ex) {
+                            throw new IllegalStateException(ex);
+                        }
+                    }
+                }
+        );
+        app.start();
+        Thread.sleep(BkBasicTest.WAIT_APP);
+        final HttpURLConnection persistConnection = HttpURLConnection
```
dmzaytsev: @mstzn36 pelase checkout this article http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#variable-names
>Avoid composite names of variables, like timeOfDay, firstItem, or httpRequest

src/test/java/org/takes/http/BkBasicTest.java
dmzaytsev: @mstzn36 same question

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +102,59 @@ public void handlesSocket() throws IOException {
...
+    public void handlesPersistentConnection() throws Exception {
+        final int port = 8080;
+        final String uri = String.format("http://localhost:%d", port);
+        final Take take = new Take() {
+            @Override
+            public Response act(final Request req) throws IOException {
+                return new TkEmpty().act(req);
+            }
+        };
+        final Thread app = new Thread(
```
dmzaytsev: @mstzn36 you can use `FtRemote` instead running BkBasic in the thread
check https://github.com/yegor256/takes/blob/master/README.md
or https://github.com/yegor256/takes/blob/master/src/test/java/org/takes/facets/auth/social/PsGithubTest.java
or any another test using `FtRemote`

dmzaytsev:
@mstzn36 4 comments above

mstzn36:
@dmzaytsev I didn't use FtRemote because i need to create a connectionthrough HttpURLConnection and manage it myself. Please see  https://github.com/yegor256/takes/pull/337#issuecomment-112151309

Other feedbacks are fixed.

dmzaytsev:
@mstzn36 got it, thanks

dmzaytsev:
@rultor merge

yegor256:
@mstzn36 there is something very wrong with indentation. Why are you using more than 4 spaces? We always (no exceptions) use 4 spaces. Please fix, it's difficult to read your code

mstzn36:
@yegor256 Sorry,I thought qulice checks all indentation standarts. I fixed

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +102,59 @@ public void handlesSocket() throws IOException {
...
 
     /**
+     * BkBasic can handle HTTP persistent connections.
+     * @throws Exception If some problem inside
+     * @checkstyle ExecutableStatementCountCheck (2 lines)
+     */
+    @Test
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    public void handlesPersistentConnection() throws Exception {
+        final int port = 8080;
```
yegor256: @mstzn36 this is not acceptable. how can you be sure that port 8080 is available on the build machine?

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +102,59 @@ public void handlesSocket() throws IOException {
...
+                    } catch (final IOException ex) {
+                        throw new IllegalStateException(ex);
+                    }
+                }
+            }
+        );
+        app.start();
+        Thread.sleep(BkBasicTest.WAIT_APP);
+        final HttpURLConnection httpURLConnection = HttpURLConnection
+                .class.cast(
```
yegor256: @mstzn36 why so many indentation spaces here?

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -87,16 +110,44 @@ public void accept(final Socket socket) throws IOException {
...
 
     /**
+     * Whether to http connection is persistent or not.
+     * @param req RqLive
+     * @return Boolean
+     * @throws IOException if fails
+     */
+    private boolean isConnectionKeepAlive(final RqLive req) throws IOException {
+        boolean keep = false;
+        final RqHeaders.Base rqHeaders = new RqHeaders.Base(req);
```
yegor256: @mstzn36 see http://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

yegor256:
@mstzn36 I posted a few comments. I still believe that implementation is wrong... I would recommend to remove it, puzzle, and leave just a test in this branch.

src/test/java/org/takes/http/BkBasicTest.java
mstzn36: @yegor256 I can detect a free port but it will increase complexity of BkBasicTest. According to your comment ( https://github.com/yegor256/takes/pull/337#issuecomment-112151309 ) I need to run the server for testing persistent connection. Do you have any suggestion ? Should i detect a free port ?

src/test/java/org/takes/http/BkBasicTest.java
mstzn36: @yegor256 It is the result of 4 space indentation.

src/main/java/org/takes/http/BkBasic.java
mstzn36: @yegor256 i renamed isConnectionKeepAlive to persistent

mstzn36:
@yegor256 I think it isn't proper for puzzle. I believe the implementation is OK but maybe there is a problem about test. We can use mock for testing if you agree. I really want to solve this issue. If you have any suggestion,i would be very happy to hear

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,16 +84,25 @@ public BkBasic(final Take tks) {
...
     }
 
+    /**
+     * Read from socket.
+     * @param socket TCP socket with HTTP
+     * @throws IOException if some problem occurs
+     * @todo #306 15min/DEV : the server should use same socket for new coming
+     *  requests if the connection is persistent.
+     *  And the server should create new socket for new coming requests if the
+     *  connection is not persistent
```
yegor256: @mstzn36 it's not clear what do you want me, a developer working with this puzzle, to do. You're just stating the fact about something... But I can't understand what needs to be done..

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -87,16 +110,44 @@ public void accept(final Socket socket) throws IOException {
...
+        final RqHeaders.Base rqHeaders = new RqHeaders.Base(req);
+        if (rqHeaders.head().iterator().hasNext()) {
+            final Iterator<String> values = rqHeaders.header(BkBasic.CONNECTION)
+                .iterator();
+            if (values.hasNext()) {
+                do {
+                    keep = BkBasic.KEEP_ALIVE.toLowerCase(
+                        Locale.ENGLISH
+                        ).equals(
+                        values.next().toLowerCase(Locale.ENGLISH)
```
yegor256: @mstzn36 `RqHeaders` returns headers already in lower case, no need to do this conversion here again

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -87,16 +110,44 @@ public void accept(final Socket socket) throws IOException {
...
     /**
+     * Whether to http connection is persistent or not.
+     * @param req RqLive
+     * @return Boolean
+     * @throws IOException if fails
+     */
+    private boolean persistent(final RqLive req) throws IOException {
+        boolean keep = false;
+        final RqHeaders.Base rqHeaders = new RqHeaders.Base(req);
+        if (rqHeaders.head().iterator().hasNext()) {
```
yegor256: @mstzn36 this is always true. I don't see why do we need this `if

src/test/java/org/takes/http/BkBasicTest.java
yegor256: @mstzn36 we can't use port 8080, it may just be not available on the build machine and the test will fail. Of course we have to allocate a random port. See how it's done in other tests

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -86,6 +102,59 @@ public void handlesSocket() throws IOException {
...
+            new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        new FtBasic(
+                            new BkBasic(take),
+                            port
+                        ).start(
+                            Exit.NEVER
+                        );
```
yegor256: @mstzn36 the formatting looks strange... can't you just format it like this:

```
new FtBasic(new BkBasic(take), port).start(Exit.NEVER);
```

if it doesn't fit in one line, I would reformat like this:

```
new FtBasic(
  new BkBasic(take), port
).start(Exit.NEVER);
```

src/test/java/org/takes/http/BkBasicTest.java
yegor256: @mstzn36 what do you mean? :) we always use 4 spaces, never 8 or more, or less. just 4.

yegor256:
@mstzn36 I'm not against you closing this ticket, but the bigger the branch, the longer we will discuss it. It's better to make the branch smaller and puzzle everything else. Anyway, see my comments above.

mstzn36:
@yegor256 thank you for comments

davvd:
@ypshenychka review this ticket please, for compliance with our [quality rules](http://www.teamed.io/qa.html)

ypshenychka:
@davvd Quality is good here.

davvd:

@ypshenychka thanks for the review

davvd:
@dmzaytsev 10 mins was added to the account of @ypshenychka (for QA review), in transaction 63437714... thanks, I just added **24 mins** to your account, payment `63437732`, 239 hours and 42 mins spent... extra minutes for review comments (c=9)... added +24 to your rating, now it is equal to [+1504](http://www.netbout.com/b/35166?open=rating)

Issue #377, opened 2015-08-21
TkProxy


pinaf:
A decorator class `TkProxy` would be useful. It would pass along requests received in `.act` to a different destination (internally handling body, headers, etc for the new request)  and also proxy back responses.

pinaf:
@yegor256 what do you think about this one?

davvd:
@yegor256 please dispatch this issue

Suseika:
@yegor256 I support the idea, too.

yegor256:
@pinaf yeah, good idea, thanks

davvd:
@pinaf milestone set to `1.0` (correct me if I am wrong)

davvd:
@pinaf thanks for this report, I added 15 mins to your account, in transaction `AP-00L05667MK3382931`

davvd:
@tiaped this ticket is yours now, please proceed, and keep in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Any technical questions you should ask right here; The budget of this task is **30 mins**. This is exactly how much time will be compensated, when the task is completed. Read about our [Definition of Done](http://www.xdsd.org/2014/04/15/definition-of-done.html)

tiaped:
Hi, I have no idea of what to put in 
 * @version $Id: 9357c24e802b82a9275a1f58d2ff0d2d4eb69397 $
 * @since 0.20

Could please help?
Are you ok with abstract classes?
I think I will make preAct and postAct methods.

tiaped:
I want also to be sure that the aim of the issue is to update request before acting and update reponse before returning.
Am I right?

tiaped:
FYI I have some issues on qulice with maven 3.0.5. Everything was fine with maven 3.3.3.

Suseika:
@tiaped Qulice doesn't work with Maven older than 3.2.5. And please address all your comments. If you're not sure to whom you should address them, read the work policy.

`$Id: 9357c24e802b82a9275a1f58d2ff0d2d4eb69397 $` is for https://stackoverflow.com/questions/384108/moving-from-cvs-to-git-id-equivalent. You basically don't touch it (it will be replaced with just `$Id$` automatically by git) and put `$Id$` in `@version` tags of your new classes.

tiaped:
@Suseika Thanks a lot

tiaped:
@davvd PR #377 is ready for review

davvd:
> @davvd PR #377 is ready for review

[...]

PR #421, opened 2015-10-30
#377: Added TkProxy class

bdragan:
#377: Added TkProxy class.


davvd:
@bdragan Let me find a reviewer for this pull request, thanks for submitting it

davvd:
@dmzaytsev please review, thanks

src/main/java/org/takes/tk/TkProxy.java
```diff
@@ -0,0 +1,144 @@
...
+ * received from the original take is returned.
+ *
+ * <p>The class is immutable and thread-safe.
+ *
+ * @author Dragan Bozanovic (bozanovicdr@gmail.com)
+ * @version $Id$
+ * @since 0.25
+ */
+@ToString
+@EqualsAndHashCode
```
dmzaytsev: @bdragan By default, @EqualsAndHashCode use all non-static, non-transient fields. So `#equals()` and `#hashCode()` will be invalid here.  You should mention transient field names in the annotation.

src/main/java/org/takes/tk/TkProxy.java
```diff
@@ -0,0 +1,144 @@
...
+            public Request transform(final Request request) {
+                return request;
+            }
+        }, rstransformr);
+    }
+
+    @Override
+    public Response act(final Request req) throws IOException {
+        final Request transformed = this.rqtransformer.transform(req);
+        final Response response = this.origin.act(transformed);
```
dmzaytsev: @bdragan `transformed` variable is redundant, it can be inline

src/main/java/org/takes/tk/TkProxy.java
```diff
@@ -0,0 +1,144 @@
...
+        return this.rstransformer.transform(response);
```
dmzaytsev: @bdragan same here, `response` can be inline

dmzaytsev:
@bdragan looks good, please see 3 my comments above
also would it possible to get rid usage of mocking objects?
see http://www.yegor256.com/2014/09/23/built-in-fake-objects.html
you can use a simple Take like TkFixed instead of mocking Take class
and just change the request body instead using of the mocking request

bdragan:
@dmzaytsev Thank you. Please see the new commit.

I got rid of mocks for `Request`s and `Response`s, but I kept the one for the original `Take`. The reason for it is that I need to verify which request is passed to it (transformed or the original one).

The alternative is to create a stub `Take` and perform assertions in its `act` method, but I find it more cumbersome than using mocks. Please feel free to suggest another approach to handle this case.

dmzaytsev:
@bdragan thank you

dmzaytsev:
@rultor merge

yegor256:
@bdragan hm... I don't think that's the point of `TkProxy` class. I don't really understand what this class may be used for... We need `TkProxy` for something like this:

```java
Take take = new TkProxy("google.com", 80);
```

Then, when I'm doing `take.act(new RqFake("/", "GET"))`, a request should go to `google.com`. Make sense?

bdragan:
@yegor256 Yes, it seems that there was a slight misunderstanding about the requirements. I'll change my implementation.

yegor256:
@bdragan this code may help you (it's not perfect, but may help you to start):

```java
public final class TkProxy implements Take {
    private final transient HostAndPort host;
    public TkProxy(final HostAndPort target) {
        this.host = target;
    }
    @Override
    public Response act(final Request req) throws IOException {
        final Request greedy = new RqGreedy(req);
        final RqHref.Base base = new RqHref.Base(greedy);
        final String dest = StringUtils.replace(
            base.href().toString(),
            new RqHref.Smart(base).home().bare(),
            String.format("http://%s/", this.host.toString())
        );
        com.jcabi.http.Request proxied = new ApacheRequest(dest)
            .method(new RqMethod.Base(greedy).method());
        final RqHeaders.Base headers = new RqHeaders.Base(greedy);
        for (final String name : headers.names()) {
            if ("content-length".equals(name.toLowerCase(Locale.ENGLISH))) {
                continue;
            }
            for (final String value : headers.header(name)) {
                proxied = proxied.header(name, value);
            }
        }
        final Collection<String> hdrs = new LinkedList<>();
        hdrs.add(
            String.format(
                "X-Xockets-TkProxy: from %s to %s",
                new RqHref.Smart(base).home().bare(), dest
            )
        );
        final byte[] body = IOUtils.toByteArray(
            new RqLengthAware(greedy).body()
        );
        final com.jcabi.http.Response rsp = proxied.fetch(
            new ByteArrayInputStream(body)
        );
        for (final Map.Entry<String, List<String>> entry
            : rsp.headers().entrySet()) {
            for (final String value : entry.getValue()) {
                final String val;
                if ("Host".equals(entry.getKey())) {
                    val = this.host.toString();
                } else {
                    val = value;
                }
                hdrs.add(String.format("%s: %s", entry.getKey(), val));
            }
        }
        return new RsWithBody(
            new RsWithHeaders(
                new RsWithStatus(new RsEmpty(), rsp.status(), rsp.reason()),
                hdrs
            ),
            rsp.binary()
        );
    }
}
```

bdragan:
@yegor256 Thanks, the example is very useful.

bdragan:
@dmzaytsev Please see the new commit.

pom.xml
```diff
@@ -152,12 +152,12 @@
         <dependency>
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
-            <scope>test</scope>
+            <optional>true</optional>
```
dmzaytsev: @bdragan I recall we should try to avoid usage of other frameworks in Takes 
it possible only in tests
@yegor256 please confirm

yegor256: @bdragan @dmzaytsev yes, that's true. there is a comment about it in `pom.xml`

bdragan:
@dmzaytsev Fixed, please see the new commit.

src/main/java/org/takes/tk/TkProxy.java
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @author Dragan Bozanovic (bozanovicdr@gmail.com)
+ * @version $Id$
+ * @since 0.25
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ * @todo #377 We need the integration test for this class.
+ *  The test should verify the different HTTP methods (GET, POST, etc),
+ *  as well as the different combinations of request/response headers.
+ */
+public class TkProxy implements Take {
```
dmzaytsev: @bdragan let's make whole class final instead a method

dmzaytsev:
@bdragan Strictly speaking, you should start from tests instead of the implementation. Teamed try to follow TDD. Otherwise how will we know know that the code works correctly?
Let's make at least one test and create a puzzle for others.
@yegor256 don't you mind?

yegor256:
@dmzaytsev not at all, TDD is the king

bdragan:
@dmzaytsev Thanks, please take a look at the new changes.

dmzaytsev:
@rultor merge

src/main/java/org/takes/tk/TkProxy.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+            if (isHost(name)) {
+                proxied = proxied.header(name, this.host);
+                continue;
+            }
+            for (final String value : headers.header(name)) {
+                proxied = proxied.header(name, value);
+            }
+        }
+        if (Arrays.asList("POST", "PUT").contains(method)) {
```
yegor256: @bdragan why only them? what about `PATCH`, for example? I would not do this filtering-out at all. All of them may have a body

src/main/java/org/takes/tk/TkProxy.java
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * Creates the response received from the target host.
+     *
+     * @param home Home host
+     * @param dest Destination URL
+     * @param rsp Response received from the target host
+     * @return Response
+     */
+    private Response response(final String home, final String dest,
+            final com.jcabi.http.Response rsp) {
```
yegor256: @bdragan what happened with indentation here?

yegor256:
@bdragan a few comments above

src/main/java/org/takes/tk/TkProxy.java
bdragan: @yegor256 Good point. Actually, I tried to implemented it that way initially, but I encountered the issue that makes all attempts to read from the body input stream from `GET` requests to block forever.

I intended to write a separate issue for that later (actually I did it now, please see #438), and I added support for proxying only POST and PUT request bodies (which should cover most of the cases in practice). However, I agree that I should have added at least a puzzle directly in this PR.

I changed my implementation to check for the presence of `Content-Length` or `Transfer-Encoding` headers before attempting to read the request body, as described [here](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3):
>The presence of a message-body in a request is signaled by the inclusion of a Content-Length or Transfer-Encoding header field in the request's message-headers.

What do you think?

src/main/java/org/takes/tk/TkProxy.java
bdragan: @yegor256 Thanks for the remarks, please see my comments and new submitted changes.

bdragan:
@yegor256 Thanks for the remarks, please see my comment and new submitted changes.

dmzaytsev:
@yegor256 take a look when you have time

yegor256:
@rultor try to merge

src/test/java/org/takes/tk/TkProxyTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.takes.http.FtRemote;
+import org.takes.rq.RqFake;
+import org.takes.rs.RsPrint;
+
+/**
+ * Test case for {@link TkProxy}.
+ * @author Dragan Bozanovic (bozanovicdr@gmail.com)
+ * @version $Id$
+ * @since 0.25
+ * @todo #377 We need more tests for TkProxy.
```
dmzaytsev: @bdragan please change to `@todo #377:30min/DEV`

bdragan:
@dmzaytsev Thanks; fixed.

dmzaytsev:
@rultor merge

yegor256:
@rultor try to merge

davvd:
@ypshenychka review this ticket please, for compliance with our [quality rules](http://at.teamed.io/qa.html)

davvd:
@rultor deploy

ypshenychka:
@davvd Quality is good here.

davvd:

@ypshenychka thanks for the review and a good remark

davvd:
@dmzaytsev added 10 mins to @ypshenychka, for QA review, payment code is `71773224`. I just added **28 mins** to your account, many thanks for your contribution (71773240).. 911 hours and 38 mins spent here. extra minutes for review comments (c=13). +28 added to your rating, current score is: [+882](http://www.netbout.com/b/35166?open=rating)

Issue #392, opened 2015-09-24
RqMethod corrects format


hs3180:
RqMethod just read first field in headers, and split it with spaces.
It should check the header format according to HTTP specification.

yegor256:
@hs3180 make sense, thanks for the catch

davvd:
@hs3180 I will ask someone to take care of this task soon

davvd:
@hs3180 attached this issue to milestone "1.0" (let me know if this is wrong)

davvd:
@hs3180 thanks for the report, I topped your acc for 15 mins, payment ID `AP-7DJ20956U6273802U`

davvd:
@bdragan it's yours, please go ahead

bdragan:
@hs3180 The current implementation of `RqMethod` already reads the method according to HTTP spec. What exactly do you have in mind? Is it validation (that it must be one of `GET`, `POST`..., exactly one space between method, URL and protocol version and ends with CRLF)?

hs3180:
@bdragan yes, let's do a validation according to http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html

bdragan:
@hs3180 Thanks, PR submitted.

hs3180:
@bdragan you should accept methods which are not in default methods. treat them as `extension-method`

bdragan:
@hs3180 Thanks for the remark. What are `extension-method`s?

hs3180:
@bdragan

       Method         = "OPTIONS"                ; Section 9.2
                      | "GET"                    ; Section 9.3
                      | "HEAD"                   ; Section 9.4
                      | "POST"                   ; Section 9.5
                      | "PUT"                    ; Section 9.6
                      | "DELETE"                 ; Section 9.7
                      | "TRACE"                  ; Section 9.8
                      | "CONNECT"                ; Section 9.9
                      | extension-method
       extension-method = token

       token          = 1*<any CHAR except CTLs or separators>

## \*rule
The character "\*" preceding an element indicates repetition. The full form is "\<n\>\*\<m\>element" indicating at least \<n\> and at most \<m\> occurrences of element. Default values are 0 and infinity so that "\*(element)" allows any number, including zero; "1\*element" requires at least one; and "1\*2element" allows one or two.

bdragan:
@hs3180 Thanks, I missed that one. New changes submitted.

bdragan:
[...]

PR #451, opened 2015-12-02
#392: RqMethod validates the HTTP request method line.

bdragan:
#392: RqMethod validates the HTTP request method line.

### DIFF
--- a/src/main/java/org/takes/rq/RqMethod.java
+++ b/src/main/java/org/takes/rq/RqMethod.java
@@ -77,6 +77,16 @@ public interface RqMethod extends Request {
     String PATCH = "PATCH";
 
     /**
+     * TRACE method.
+     */
+    String TRACE = "TRACE";
+
+    /**
+     * CONNECT method.
+     */
+    String CONNECT = "CONNECT";
+
+    /**
      * Get method.
      * @return HTTP method
      * @throws IOException If fails
@@ -96,10 +106,18 @@ public interface RqMethod extends Request {
     final class Base extends RqWrap implements RqMethod {
 
         /**
+         * HTTP token separatos which are not already excluded by PATTERN.
+         */
+        private static final Pattern SEPARATORS = Pattern.compile(
+            "[()<>@,;:\\\"/\\[\\]?={}]"
+        );
+
+        /**
          * HTTP method line pattern.
+         * [!-~] is for method or extension-method token (octets 33 - 126).
          */
         private static final Pattern PATTERN = Pattern.compile(
-            "(GET|POST|PUT|DELETE|HEAD|OPTIONS|PATCH) [^ ]+( [^ ]+){0,1}"
+            "([!-~]+) [^ ]+( [^ ]+){0,1}"
         );
 
         /**
@@ -117,8 +135,11 @@ public interface RqMethod extends Request {
             if (!matcher.matches()) {
                 throw new InvalidHTTPMethodLineException(line);
             }
-            final String[] parts = line.split(" ", 2);
-            return parts[0].toUpperCase(Locale.ENGLISH);
+            final String method = matcher.group(1);
+            if (SEPARATORS.matcher(method).find()) {
+                throw new InvalidHTTPMethodLineException(line);
+            }
+            return method.toUpperCase(Locale.ENGLISH);
         }
     }
 
--- a/src/main/java/org/takes/rq/RqMethod.java
+++ b/src/main/java/org/takes/rq/RqMethod.java
@@ -133,32 +133,24 @@ public interface RqMethod extends Request {
             final String line = this.head().iterator().next();
             final Matcher matcher = PATTERN.matcher(line);
             if (!matcher.matches()) {
-                throw new InvalidHTTPMethodLineException(line);
+                throw invalidHttpMethodLine(line);
             }
             final String method = matcher.group(1);
             if (SEPARATORS.matcher(method).find()) {
-                throw new InvalidHTTPMethodLineException(line);
+                throw invalidHttpMethodLine(line);
             }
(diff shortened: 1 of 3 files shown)


davvd:
@bdragan Thanks, let me find someone who can review this pull request

davvd:
@dmzaytsev it's yours, please review

src/main/java/org/takes/rq/RqMethod.java
```diff
@@ -94,6 +96,13 @@
     final class Base extends RqWrap implements RqMethod {
 
         /**
+         * HTTP method line pattern.
+         */
+        private static final Pattern PATTERN = Pattern.compile(
+            "(GET|POST|PUT|DELETE|HEAD|OPTIONS|PATCH) [^ ]+( [^ ]+){0,1}"
```
dmzaytsev: @bdragan what about extension-method that the ticket reporter mentioned?

src/test/java/org/takes/rq/RqMethodTest.java
```diff
@@ -47,4 +49,54 @@ public void returnsMethod() throws IOException {
...
+     * RqMethod can fail when HTTP method line has any extra undefined
+     * elements.
+     * @throws IOException If some problem inside
+     */
+    @Test(expected = InvalidHTTPMethodLineException.class)
+    public void failsOnExtraLineElement() throws IOException {
+        new RqMethod.Base(
+            new RqSimple(Arrays.asList("GET / HTTP/1.1 abc"), null)
+        ).method();
+    }
```
dmzaytsev: @bdragan lets add a new test for the extension-method

src/test/java/org/takes/rq/RqMethodTest.java
```diff
@@ -47,4 +49,54 @@ public void returnsMethod() throws IOException {
...
+     * RqMethod supports all HTTP methods.
+     * @throws IOException If some problem inside
+     */
+    @Test
+    @SuppressWarnings("PMD.AvoidInstantiatingObjectsInLoops")
+    public void supportsAllMethods() throws IOException {
+        for (final String method
+            : Arrays.asList(
+                RqMethod.DELETE, RqMethod.GET, RqMethod.HEAD, RqMethod.OPTIONS,
+                RqMethod.PATCH, RqMethod.POST, RqMethod.PUT
```
dmzaytsev: @bdragan there is method `TRACE` in [RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html)

src/main/java/org/takes/rq/RqMethod.java
dmzaytsev: @bdragan there is only one space allows after uri (http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html)
```
Request-Line   = Method SP Request-URI SP HTTP-Version CRLF
```
but the method line like the string below will be valid
```
GET http://www.w3.org/pub/WWW/TheProject.html                               HTTP/1.1
```

dmzaytsev:
@bdragan a few comments above

src/main/java/org/takes/rq/RqMethod.java
bdragan: @dmzaytsev I allow only one space in the regex part after the URI group of characters: ( [^ ]+)
All of the other characters must be non-space ones: [^ ]

bdragan:
@dmzaytsev Thanks, please see the new changes.

dmzaytsev:
@bdragan thanks

dmzaytsev:
@rultor merge

src/main/java/org/takes/rq/RqMethod.java
```diff
@@ -104,8 +131,34 @@ public Base(final Request req) {
         @Override
         public String method() throws IOException {
             final String line = this.head().iterator().next();
-            final String[] parts = line.split(" ", 2);
-            return parts[0].toUpperCase(Locale.ENGLISH);
+            final Matcher matcher = PATTERN.matcher(line);
+            if (!matcher.matches()) {
+                throw new InvalidHTTPMethodLineException(line);
```
yegor256: @bdragan what is the point? why can't we just throw `IOException`?

yegor256:
@bdragan this exception class is redundant, just `IOException` is enough, see http://www.yegor256.com/2015/07/28/checked-vs-unchecked-exceptions.html

bdragan:
@yegor256 Thanks; fixed.

yegor256:
@bdragan I understand why you're introducing this supplementary method - to avoid code duplication. But it's a bad idea in this case. Don't be afraid of code duplication, if your code is doing different things. First, it throws an exception if something is wrong. Then, it throws if something **else** is wrong. There is **nothing** in common between these two situations. Get rid of that static method.

bdragan:
@yegor256 True, in this case it is not needed; even the message should be a bit different. Thanks; fixed.

yegor256:
@rultor try to merge

davvd:
@ypshenychka please, help us to evaluate quality here, as per [par.24](http://at.teamed.io/policy.html#24)

davvd:
@rultor deploy now pls

ypshenychka:
@davvd Quality is good.

davvd:
> @davvd Quality is good.
> 

@ypshenychka thank you, good quality is always good ;)

davvd:
@dmzaytsev added 10 mins to @ypshenychka, for QA review, payment code is `71781473`; Much obliged! I have added **22 mins** to your account in payment "71781502", 121 hours and 23 mins spent; you're getting extra minutes here (c=7); added +22 to your rating, now it is equal to [+904](http://www.netbout.com/b/35166?open=rating)

Issue #442, opened 2015-11-17
inputStream in RsVelocity.render(InputStream, Map<CharSequence, Object>) never closes


simonjenga:

davvd:
@yegor256 dispatch this issue please, see [par.21](http://at.teamed.io/policy.html#21)

yegor256:
@simonjenga seems to be a valid bug, thanks

davvd:
@simonjenga attached this issue to milestone "1.0" (let me know if this is wrong)

davvd:
@simonjenga many thanks for the bug report, I added 15 mins to your acc, transaction `71680633`

gauee:
Can I start working on this issue?

davvd:
@happy-neko please pick this up, and keep in mind [these instructions](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Any technical questions - ask right here. Task's budget is **30 mins** (see [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for explanation)

Happy-Neko:
@simonjenga PR #442 can be merged to close this ticket.

simonjenga:
@Happy-Neko ```inputStream``` in ```RsVelocity.render(InputStream, Map<CharSequence, Object>)``` is not yet closed

simonjenga:
@Happy-Neko While i also like the design and your ideas, i expected the fix to be simply ```input.close();```. Always try to keep code to be as minimum as possible so as to avoid bloatware and future maintenance nightmares.

I also think we should also add ```input.close();``` in ```RsVelocity.render(InputStream, Map<CharSequence, Object>)``` to avoid somebody else raising a similar ticket.
Thanks!

Happy-Neko:
@simonjenga Please see PR description. InputStream is already closed in template engine, there is no need to close it twice.

simonjenga:
@Happy-Neko OK!

davvd:
@ypshenychka please, review this task for compliance with our [quality rules](http://at.teamed.io/qa.html)

ypshenychka:
@davvd Quality is good here.

davvd:
> @davvd Quality is good here.
> 

@ypshenychka thanks a lot

davvd:
[...]

PR #465, opened 2015-12-14
#442: RsVelocity.render() should close template's InputStream

Happy-Neko:
Could not reproduce issue #442 with unit tests. It appears that Velocity engine closes template's stream internally.

@david Please assign someone to do code review.


davvd:
@happy-neko Thanks, someone will review your pull request soon

davvd:
@carlosmiranda review this one, please

src/test/java/org/takes/misc/MockInputStream.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Converts supplied string to an input stream encoded as bytes in default
+ * encoding.
+ *
+ * @author I. Sokolov (happy.neko@gmail.com)
+ * @version $Id$
+ */
+public final class MockInputStream extends ByteArrayInputStream {
```
carlosmiranda: @Happy-Neko Why do we need this class? It is very problematic. It's named as if it were a mock but is actually doing String encoding. Then it extends `ByteArrayInputStream`, but only initializes the internal array to a string value. No need for a separate class for that.

src/test/java/org/takes/misc/MockInputStream.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+
+    /**
+     * Instantiates MockInputStream based on supplied string. String is
+     * converted to bytes stream using default charset encoding.
+     * @param input Source String
+     */
+    public MockInputStream(final String input) {
+        super(input.getBytes());
```
carlosmiranda: @Happy-Neko Default encoding... not a good idea. Better to use `UTF-8` unless there's a good reason.

src/test/java/org/takes/rs/RsVelocityTest.java
```diff
@@ -54,4 +56,23 @@ public void buildsTextResponse() throws IOException {
...
+        MatcherAssert.assertThat(
+            IOUtils.toString(
+                new RsVelocity(
+                    templateStream,
+                    Collections.<CharSequence, Object>emptyMap()
+                ).body()
+            ),
+            Matchers.equalTo(template)
+        );
+        MatcherAssert.assertThat(templateStream.isClosed(), Matchers.is(true));
```
carlosmiranda: @Happy-Neko Better to mock `InputStream` with Mockito here, since it's the only behavior we're testing.

src/test/java/org/takes/misc/MockInputStream.java
Happy-Neko: @carlosmiranda Default encoding is used deliberately. README.md states that:

> The entire framework relies on your default Java encoding, which is not necessarily UTF-8 by default. To be sure, always set it on the command line with file.encoding Java argument. We decided not to hard-code "UTF-8" in our code mostly because this would be against the entire idea of Java localization, according to which a user always should have a choice of encoding and language selection. We're using Charset.defaultCharset() everywhere in the code.

and specifically

> If your default encoding is not UTF-8, some of unit tests will break. This is an intentional behavior.

carlosmiranda: @Happy-Neko Thanks for the explanation. I'm still skeptical about the whole class itself, seems too much for this test

src/test/java/org/takes/misc/MockInputStream.java
Happy-Neko: @carlosmiranda We need to test if particular `InputStream` was closed or not. `ByteArrayInputStream` was chosen as base class because it allows to easily feed test data without reimplementing `InputStream` from scratch. 

Maybe `MockInputStream` should be renamed to something more clear like `StateAwareInputStream`.

src/test/java/org/takes/rs/RsVelocityTest.java
Happy-Neko: @carlosmiranda As you can see here https://github.com/Happy-Neko/takes/blob/49f4c47b87a3c9a5ae0e2fec343f1ebee6b01798/src/main/java/org/takes/rs/RsVelocity.java#L146 `InputStream` under test is passed to external library (Velocity) and we want to test whether it was closed here or not. I don't think it will be easy to mock it with Mockito since we have no control on what methods are called within external code. May be it is possible to use partial mocks to spy on real `InputStream` object. But extending real `InputStream` with required behavior feels like cleaner solution.

src/test/java/org/takes/misc/MockInputStream.java
carlosmiranda: @Happy-Neko Understood. In that case:

1. Let's remove the string encoding logic, it's unnecessary.
2. We should create a [Decorator](http://www.yegor256.com/2015/02/26/composable-decorators.html) of `InputStream` instead of extending `ByteArrayInputStream`.

Happy-Neko:
@carlosmiranda Please look at my 3 comments above and share your thoughts. 

Just a quick summery of what I want to achieve with unit test: make sure that template stream passed to `RsVelocity` is properly closed inside Velocity engine that is called internally.

carlosmiranda:
@Happy-Neko Again, let's create a `Decorator` of `InputStream` instead. Pass a `ByteArrayInputStream` to it. No need to overload the class with the string decoding logic, let's do that in the test itself.

Happy-Neko:
@carlosmiranda refactoring done.

carlosmiranda:
@Happy-Neko Thanks!

carlosmiranda:
@rultor merge please

Happy-Neko:
@yegor256 ping

yegor256:
@rultor try to merge

carlosmiranda:
@Happy-Neko There are merge conflicts on this branch, please resync with master.

Happy-Neko:
@rultor try merge

Happy-Neko:
@yegor256 please confirm merge.

Happy-Neko:
> @happy-neko Thanks for your request. @yegor256 Please confirm this.

@yegor256 Please respond.

carlosmiranda:
@rultor merge please

carlosmiranda:
@yegor256 ping!

yegor256:
@rultor merge

davvd:
@elenavolokhova review this ticket please, for compliance with our [quality rules](http://at.teamed.io/qa.html)

davvd:
@rultor please deploy

elenavolokhova:
@davvd Quality is good here.

davvd:

@elenavolokhova thanks, got it, great to see high-quality tickets here

davvd:
@carlosmiranda added 10 mins to @elenavolokhova, for QA review, payment code is `74654340`

I added **26 mins** to your account, many thanks for working with the project (74654347)! The completion time here was 721 hours and 23 mins.

you're getting extra minutes here (c=11)

added +26 to your rating, now it is equal to [+2577](http://www.netbout.com/b/35023?open=rating)

Issue #519, opened 2016-01-16
BkBasicTest.java:57-60: At the moment we don't...


davvd:
Puzzle `306-b987400a` in `src/test/java/org/takes/http/BkBasicTest.java:57-60` has to be resolved: At the moment we don't support HTTP persistent connections. Would be great to implement this feature. BkBasic.accept should handle more than one HTTP request in one connection.... The puzzle was created by Maksimenko Vladimir on 11-Jan-2016.

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is removed from the source code.

<hr id='w'>

davvd:
@ddaanniieell01 the task is for you now, follow [these guidelines](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). Don't hesitate to ask any technical questions right here... This task's budget is **30 mins**. This is exactly how much will be paid when the problem explained above is solved. See [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) for more information

davvd:
@yegor256 there is no milestone yet, so I set it to `1.0`

davvd:
@exper0 it's yours,please go ahead

exper0:
@davvd need more time here

davvd:
> @davvd need more time here

@exper0 OK, take your time

exper0:
@davvd please assign someone else.

davvd:
> @davvd please assign someone else.

@exper0 I deducted 30 points from your rating

davvd:
> @davvd please assign someone else.

@exper0 I will try to find someone else

davvd:
@sebing the task is yours, please go ahead

sebing:
@yegor256  I need some design help.
The problem is `org.takes.rq.RqLive#parse` hiding the `Inputstream` and creating a `ByteArrayOutputStream` So the the read in `org.takes.rs.RsPrint#printBody(java.io.OutputStream)` is happening in the new object not in the socket's Inputstream. 

Hope you understand me ;)

1) I need some guidance on why we are hiding the Inputstream.
2) Should I expose the `org.takes.rq.RqLive#data` so In `org.takes.http.BkBasic#accept` I can know the read availability by checking something like this `if (data.has())`  

Did I understand the spaghetti code? May be my understanding of the code is not enough!

sebing:
[...]

PR #569, opened 2016-02-04
#519 - persistent HTTP connection

exper0:
#519 - added ability to process more than one request in same connection. Existing test `handlesTwoRequestInOneConnection` is now passing. In order to enable this processing I had to add body consumption in TkText (since it doesn't use request body at all preventing from handling next request)

### DIFF
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -23,6 +23,7 @@
  */
 package org.takes.http;
 
+import com.jcabi.aspects.Tv;
 import java.io.BufferedOutputStream;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
@@ -32,12 +33,16 @@ import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.net.HttpURLConnection;
 import java.net.Socket;
+import java.net.SocketTimeoutException;
 import lombok.EqualsAndHashCode;
 import org.takes.HttpException;
 import org.takes.Request;
 import org.takes.Response;
 import org.takes.Take;
+import org.takes.rq.RqHeaders;
+import org.takes.rq.RqLengthAware;
 import org.takes.rq.RqLive;
+import org.takes.rq.RqPrint;
 import org.takes.rq.RqWithHeaders;
 import org.takes.rs.RsPrint;
 import org.takes.rs.RsText;
@@ -58,6 +63,11 @@ import org.takes.rs.RsWithStatus;
 public final class BkBasic implements Back {
 
     /**
+     * Default socket timeouts (milliseconds).
+     */
+    private static final int TIMEOUT = Tv.THOUSAND;
+
+    /**
      * Take.
      */
     private final transient Take take;
@@ -70,17 +80,41 @@ public final class BkBasic implements Back {
         this.take = tks;
     }
 
+//      @todo #519:30min Need to support Keep-Alive header here and take
+//       into account specified timeout and max requests.
+//       See http://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html
+//       for details
     @Override
+    @SuppressWarnings({
+        "PMD.EmptyCatchBlock",
+        "PMD.AvoidInstantiatingObjectsInLoops"})
     public void accept(final Socket socket) throws IOException {
+        socket.setSoTimeout(BkBasic.TIMEOUT);
         final InputStream input = socket.getInputStream();
         try {
-            this.print(
-                BkBasic.addSocketHeaders(
-                    new RqLive(input),
-                    socket
-                ),
-                new BufferedOutputStream(socket.getOutputStream())
-            );
+            while (true) {
+                final Request request = new RqLengthAware(new RqLive(input));
+                this.print(
+                    BkBasic.addSocketHeaders(
+                        request,
+                        socket
+                    ),
+                    new BufferedOutputStream(socket.getOutputStream())
+                );
+                if (new RqHeaders.Base(request)
+                        .header("Content-Length")
+                        .size() > 0) {
+                    new RqPrint(request).printBody();
+                    input.read();
+                    input.read();
+                    if (input.available() <= 0) {
+                        break;
+                    }
+                } else {
+                    break;
+                }
+            }
+        } catch (final SocketTimeoutException exc) {
         } finally {
             input.close();
         }
@@ -108,7 +142,7 @@ public final class BkBasic implements Back {
                 )
             ).print(output);
         } finally {
-            output.close();
+            output.flush();
         }
     }
 
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -55,6 +55,10 @@ import org.takes.rs.RsWithStatus;
  * @since 0.1
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  * @checkstyle IndentationCheck (500 lines)
+ * @todo #519:30min The logic of persistent connection should be moved to
+ *  separate implementation of Back. BkBasic should be refactored as well in
+ *  order to remove connection management (otherwise new Back implementation
+ *  cannot be used as decorator for BkBasic)
  */
 @EqualsAndHashCode(of = "take")
 public final class BkBasic implements Back {
@@ -104,6 +108,11 @@ public final class BkBasic implements Back {
                 }
             }
         } catch (final SocketTimeoutException exc) {
+            // @checkstyle MethodBodyCommentsCheck (4 lines)
+            // This exception is thrown on socket timeout, this is just
+            // indicator that no more request can be handled in this connection.
+            // No need to throw it upper and no need to do anything specific
+            // on this exception.
         } finally {
             input.close();
         }
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -77,15 +77,14 @@ public final class BkBasic implements Back {
         this.take = tks;
     }
 
-//      @todo #519:30min Need to support Keep-Alive header here and take
-//       into account specified timeout and max requests parameters.
-//       See http://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html
-//       for details. Also need to remove input.available <= 0 condition
-//       - this is done in order to prevent failure of some tests,
-//       should be replaced with proper Keep-Alive header in tests
+    // @todo #519:30min Need to support Keep-Alive header here and take
+    //  into account specified timeout and max requests parameters.
+    //  See http://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html
+    //  for details. Also need to remove input.available <= 0 condition
+    //  - this is done in order to prevent failure of some tests,
+    //  should be replaced with proper Keep-Alive header in tests
     @Override
-    @SuppressWarnings({
-        "PMD.EmptyCatchBlock",
+    @SuppressWarnings({"PMD.EmptyCatchBlock",
         "PMD.AvoidInstantiatingObjectsInLoops"})
     public void accept(final Socket socket) throws IOException {
         socket.setSoTimeout(BkBasic.TIMEOUT);
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -64,11 +64,6 @@ import org.takes.rs.RsWithStatus;
 public final class BkBasic implements Back {
 
     /**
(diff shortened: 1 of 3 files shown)


davvd:
@exper0 Thanks for the pull request, let me find a reviewer..

davvd:
@krzyk please review, thanks

krzyk:
@exper0 could you write a bit more what you did in the description? Title shouldn't be the same as description.

exper0:
@krzyk see updated description.

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,17 +77,34 @@ public BkBasic(final Take tks) {
         this.take = tks;
     }
 
+//      @todo #519:30min Need to support Keep-Alive header here and take
```
krzyk: @exper0 you have to big indentation inside the comment, `//` should be at the same level as the method, and the comment should start after a single space (todo continuation should start at two spaces)

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,17 +77,34 @@ public BkBasic(final Take tks) {
...
+//       into account specified timeout and max requests parameters.
+//       See http://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html
+//       for details. Also need to remove input.available <= 0 condition
+//       - this is done in order to prevent failure of some tests,
+//       should be replaced with proper Keep-Alive header in tests
     @Override
+    @SuppressWarnings({
```
krzyk: @exper0 annotation should be formatted (at least with current qulice) like:
```
@SuppressWarnings({ "PMD.EmptyCatchBlock", "PMD.AvoidInstantiatingObjectsInLoops" })
```
or, if the line length is to big:
```
@SuppressWarnings({ "PMD.EmptyCatchBlock",
    "PMD.AvoidInstantiatingObjectsInLoops" })
```

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,17 +77,34 @@ public BkBasic(final Take tks) {
...
+                        request,
+                        socket
+                    ),
+                    new BufferedOutputStream(socket.getOutputStream())
+                );
+                if (input.available() <= 0) {
+                    break;
+                }
+            }
+        } catch (final SocketTimeoutException exc) {
```
krzyk: @exper0 why empty?

src/main/java/org/takes/tk/TkText.java
```diff
@@ -64,7 +66,13 @@ public TkText(final String body) {
         super(
             new Take() {
                 @Override
+                @SuppressWarnings("PMD.AvoidThrowingRawExceptionTypes")
                 public Response act(final Request req) {
+                    try {
+                        new RqPrint(req).printBody();
+                    } catch (final IOException exc) {
+                        throw new RuntimeException(exc);
```
krzyk: @exper0 throw `IllegalStateException` instead, and remove the suppress.

src/main/java/org/takes/tk/TkText.java
```diff
@@ -79,7 +87,13 @@ public TkText(final byte[] body) {
...
```
krzyk: @exper0 same as above, use `IllegalStateException` and remove the suppress

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -146,7 +143,13 @@ public void addressesInHeadersAddedWithoutSlashes() throws IOException {
         new BkBasic(
             new Take() {
                 @Override
+                @SuppressWarnings("PMD.AvoidThrowingRawExceptionTypes")
                 public Response act(final Request req) {
+                    try {
+                        new RqPrint(req).printBody();
+                    } catch (final IOException exc) {
+                        throw new RuntimeException(exc);
```
krzyk: @exper0 and again, `IllegalStateException` and remove suppress

src/main/java/org/takes/http/BkBasic.java
exper0: @krzyk we are reusing the same connection for a limited time - if time is over we just need to close the socket. No need to do anything specific on timeout, at the same time it's not an error(no need to throw it on upper level)

krzyk:
@exper0 I think that `BkBasic` should stay the same, and the logic of handling persistent connection should be done in a separate `Back`, ideally a decorator of a `Back` (that would need some clever coding).
Right now BkBasic is not basic anymore :)

krzyk:
@exper0 and there is one test failing in travis

exper0:
@krzyk BkBasic now (in master branch) handles just one request and closes connection. So it's impossible to write decorator for BkBasic handling multiple requests in same connection. It would be a parallel implementation of `Back` interface that needs to be explicitly used everywhere instead of BkBasic (as a replacement rather than decorator).

exper0:
@krzyk the test is failing only on JDK8 build, it's pretty weird. Will try to reproduce it (normally I use JDK7)

exper0:
@krzyk implementing the logic of persistent connection in separate decorator `Back` requires some deep refactoring, decoupling connection management and request handling (so BkBasic should be even more basic than now, and connections should be controlled somewhere else). Probably it's worth creating one more puzzle, what do you think?

krzyk:
@exper0 yes, it is a bigger task, so puzzle is a good start

src/main/java/org/takes/http/BkBasic.java
krzyk: @exper0 please add a comment in this block describing why it is empty, after that you should be able to remove `EmptyCatchBlock` exclude above

exper0:
@krzyk added description of empty block and added one more puzzle. Please take a look

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,17 +81,38 @@ public BkBasic(final Take tks) {
...
     }
 
+    // @todo #519:30min Need to support Keep-Alive header here and take
+    //  into account specified timeout and max requests parameters.
+    //  See http://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html
+    //  for details. Also need to remove input.available <= 0 condition
+    //  - this is done in order to prevent failure of some tests,
+    //  should be replaced with proper Keep-Alive header in tests
     @Override
+    @SuppressWarnings({"PMD.EmptyCatchBlock",
```
krzyk: @exper0 this shouldn't be needed now

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -53,11 +55,20 @@
...
+ *  order to remove connection management (otherwise new Back implementation
+ *  cannot be used as decorator for BkBasic)
  */
 @EqualsAndHashCode(of = "take")
 public final class BkBasic implements Back {
 
     /**
+     * Default socket timeouts (milliseconds).
+     */
+    private static final int TIMEOUT = Tv.THOUSAND;
```
krzyk: @exper0 this is u

krzyk: @exper0 this is used only once, please inline it

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -108,7 +140,7 @@ private void print(final Request req, final OutputStream output)
                 )
             ).print(output);
         } finally {
-            output.close();
+            output.flush();
```
krzyk: @exper0 if you remove class sing here you should do it a some point in accept method

krzyk:
@exper0 see my additional comments

src/main/java/org/takes/http/BkBasic.java
exper0: @krzyk PMD complains on this even with comment inside catch block

exper0:
@krzyk done. PMD EmptyCatchBlock suppression is still there because removing it fails the build (PMD expect something more than just comment in catch block)

krzyk:
@exper0 OK, I see that we don't have such workaround enabled in qulice.

krzyk:
@rultor merge pls

src/main/java/org/takes/tk/TkText.java
```diff
@@ -65,6 +67,11 @@ public TkText(final String body) {
             new Take() {
                 @Override
                 public Response act(final Request req) {
+                    try {
+                        new RqPrint(req).printBody();
+                    } catch (final IOException exc) {
```
yegor256: @exper0 why catching here? you can just let it float up

src/main/java/org/takes/tk/TkText.java
```diff
@@ -80,6 +87,11 @@ public TkText(final byte[] body) {
...
```
yegor256: @exper0 why catching here? you can just let it float up

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -147,6 +144,11 @@ public void addressesInHeadersAddedWithoutSlashes() throws IOException {
             new Take() {
                 @Override
                 public Response act(final Request req) {
+                    try {
+                        new RqPrint(req).printBody();
+                    } catch (final IOException exc) {
```
yegor256: @exper0 why catching here? you can just let it float up

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -70,19 +76,43 @@ public BkBasic(final Take tks) {
...
+    //  into account specified timeout and max requests parameters.
+    //  See http://tools.ietf.org/id/draft-thomson-hybi-http-timeout-01.html
+    //  for details. Also need to remove input.available <= 0 condition
+    //  - this is done in order to prevent failure of some tests,
+    //  should be replaced with proper Keep-Alive header in tests
     @Override
+    @SuppressWarnings({"PMD.EmptyCatchBlock",
+        "PMD.AvoidInstantiatingObjectsInLoops"})
     public void accept(final Socket socket) throws IOException {
+        socket.setSoTimeout(Tv.THOUSAND);
```
yegor256: @exper0 why 1000?

yegor256:
@exper0 a few comments above... besides that, this game you're playing with timeouts does smell... a timeout is a timeout, we should not rely on that and turn that exceptions into logical triggers. it's just not how exceptions work.

src/main/java/org/takes/tk/TkText.java
exper0: @yegor256 IOException is a checked exception. Signature `public Response act(final Request req)` doesn't contain `throws IOException` so I cannot to let it float up

src/main/java/org/takes/tk/TkText.java
exper0: @yegor256 same here, see my comment above

src/test/java/org/takes/http/BkBasicTest.java
exper0: @yegor256 same here

src/main/java/org/takes/http/BkBasic.java
exper0: @yegor256 I think it's reasonable temporary value(1 second) at this stage. Once we have Keep-Alive header support with right `max` and `timeout` we can review this default value. Having too big value will lead to very long build (because each test will be waiting given number of seconds)

exper0:
@yegor256 see my comments above. What's wrong with timeouts? Persistent HTTP connection assumes timeouts anyway - we should keep connection for a limited time. Since we use blocking read natural way to stop it is to use timeout on connection(i.e. socket). Do you agree? If not please share your thoughts on how will should work instead

exper0:
@yegor256 ping

exper0:
@yegor256 please return to this PR - I answered to your questions

yegor256:
@exper0 I don't even want to spend time to understand what is done here. this code is not acceptable:

```
} catch (final SocketTimeoutException exc) {
}
```

Just no, under **no** conditions.

yegor256:
@krzyk how come you accepted this? catching an exception (an important one) and then swallowing it??

krzyk:
@yegor256 sorry for that :(

exper0:
@yegor256 ok, I'll give a chance to someone else to guess what would you like see here.

davvd:
@ypshenychka please, let us know what do you think about this ticket, according to our [QA rules](http://at.teamed.io/qa.html)

ypshenychka:
@davvd Quality is good here.

davvd:

@ypshenychka many thanks

davvd:
@krzyk paied 10 mins to @ypshenychka for QA review (payment ID is `78674979`). paid, thanks, added **33 mins** to your account, payment `78674989`, 528 hours was spent. extra minutes for review comments (c=18). +33 to your rating, your total score is [+5619](http://www.netbout.com/b/35035?open=rating)

Issue #524, opened 2016-01-18
many implementations of org.takes.Response#body ignore charset


rfqu:
HTTP response header usually has content-type header value with definition of charset. This charset must be taken into account when characters of the body are converted to bytes using the body() method.
But most (if not all) implementations of this method uses default charset, not that defined in the header.
This can happen indirectly, like in constructor 

```
public  RsWithBody(final Response res, final CharSequence body) {
        this(res, body.toString().getBytes());
    }
```

Here getBytes() use default charset instead of the charset defined in the header.

As a result, client which receives byte stream cannot reconstruct the characters of the response body.

davvd:
@yegor256 please pay attention to this issue ([par.21](http://at.teamed.io/policy.html#21))

davvd:
@rfqu I added milestone `1.0` to this issue, let me know if there has to be something else

davvd:
@alexey-krylov this task is for you, keep [these principles](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) in mind, and don't hesitate to ask any technical questions

The cost of this task is **30 mins** (this is exactly how much will be paid, not less not more), when the task is [done](http://www.xdsd.org/2014/04/15/definition-of-done.html)

davvd:
@rfqu many thanks for the bug report, I added 15 mins to your acc, transaction `AP-4TS36903P0225481L`

davvd:
@essobedo can you please help? Keep in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). If you have any technical questions, don't hesitate to ask right here. Budget here is **30 mins** (keep [this principle](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html) in mind)

essobedo:
@rfqu Here is my pull request https://github.com/yegor256/takes/pull/565, please check it and if it is fine with you, close the ticket

rfqu:
pull request #565 is ok

essobedo:
@rfqu sorry my mistake this ticket should not be closed as long as the PR has not been merged, could you please reopen it?

essobedo:
@rfqu thx

essobedo:
@mkordas thank you for your remarks, here is my new PR https://github.com/yegor256/takes/pull/571

mkordas:
@rfqu 
> pull request #565 is ok

Please update your comment with a name tag.

rfqu:
pull request #571 is ok

[...]

PR #571, opened 2016-02-05
#524: Added the ability to specify the charset if needed

essobedo:
For #524 

The goal of this PR is to add the charset into the content type header by using the new constructor of RsWithType allowing to explicitly specify the charset.

### DIFF
--- a/src/main/java/org/takes/rs/RsWithBody.java
+++ b/src/main/java/org/takes/rs/RsWithBody.java
@@ -52,7 +52,8 @@ import org.takes.Response;
 public final class RsWithBody extends RsWrap {
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the specified body that will be
+     * encoded into UTF-8 by default.
      * @param body Body
      */
     public RsWithBody(final CharSequence body) {
@@ -60,7 +61,7 @@ public final class RsWithBody extends RsWrap {
     }
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the specified body.
      * @param body Body
      */
     public RsWithBody(final byte[] body) {
@@ -68,7 +69,7 @@ public final class RsWithBody extends RsWrap {
     }
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the specified body.
      * @param body Body
      */
     public RsWithBody(final InputStream body) {
@@ -76,7 +77,8 @@ public final class RsWithBody extends RsWrap {
     }
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the content located at the specified
+     * url as body.
      * @param url URL with body
      */
     public RsWithBody(final URL url) {
@@ -84,7 +86,8 @@ public final class RsWithBody extends RsWrap {
     }
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the specified response and body. The
+     * body will be encoded into UTF-8 by default.
      * @param res Original response
      * @param body Body
      */
@@ -93,7 +96,8 @@ public final class RsWithBody extends RsWrap {
     }
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the specified response and body. The
+     * body will be encoded using the specified character set.
      * @param res Original response
      * @param body Body
      * @param charset The character set to use to serialize the body
--- a/src/main/java/org/takes/rs/RsWithBody.java
+++ b/src/main/java/org/takes/rs/RsWithBody.java
@@ -27,6 +27,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
+import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
@@ -88,7 +89,18 @@ public final class RsWithBody extends RsWrap {
      * @param body Body
      */
     public RsWithBody(final Response res, final CharSequence body) {
-        this(res, body.toString().getBytes(StandardCharsets.UTF_8));
+        this(res, body, StandardCharsets.UTF_8);
+    }
+
+    /**
+     * Ctor.
+     * @param res Original response
+     * @param body Body
+     * @param charset The character set to use to serialize the body
+     */
+    public RsWithBody(final Response res, final CharSequence body,
+        final Charset charset) {
+        this(res, body.toString().getBytes(charset));
     }
 
     /**
--- a/src/main/java/org/takes/rs/RsWithType.java
+++ b/src/main/java/org/takes/rs/RsWithType.java
@@ -159,7 +159,7 @@ public final class RsWithType extends RsWrap {
         /**
          * Constructs a {@code HTML} that will add text/html as the content type
          * header to the response using the specified charset as charset
-         * parameter value if present..
+         * parameter value if present.
          * @param res Original response
          * @param charset The character set to add in the content type header if
          *  present.
--- a/src/main/java/org/takes/rs/RsWithType.java
+++ b/src/main/java/org/takes/rs/RsWithType.java
@@ -27,6 +27,7 @@ import java.nio.charset.Charset;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import org.takes.Response;
+import org.takes.misc.Opt;
 
 /**
  * Response decorator, with content type.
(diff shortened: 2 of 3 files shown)


davvd:
@essobedo Thanks for the pull request, let me find a reviewer..

davvd:
@mkordas review this one, please

mkordas:
@essobedo I'm on it

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -83,7 +137,16 @@ private static Response make(final Response res, final CharSequence type) {
          * @param res Original response
          */
         public HTML(final Response res) {
-            super(RsWithType.make(res, "text/html"));
+            super(RsWithType.make(res, RsWithType.TYPE_HTML));
```
mkordas: @essobedo can't we invoke `this(res, StandardCharsets.UTF_8)` here?

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -46,6 +47,31 @@
...
+
+    /**
+     * Content type text/plain.
+     */
+    private static final String TYPE_TEXT = "text/plain";
+
+    /**
+     * Content type application/json.
+     */
+    private static final String TYPE_JSON = "application/json";
```
mkordas: @essobedo can we somehow avoid the code duplication and remove these constants?

src/test/java/org/takes/rs/RsWithTypeTest.java
```diff
@@ -201,21 +256,77 @@ public void replacesTypeWithXml() throws Exception {
...
+                    ""
+                )
+            )
+        );
+    }
+
+    /**
+     * RsWithType can add the charset to the content type when it is explicitly
+     * specified and can still provide the content type without the charset
+     * when it is not explicitly specified.
```
mkordas: @essobedo are you sure that we need one test to test two things (content with charset and without charset)?

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -58,6 +84,17 @@ public RsWithType(final Response res, final CharSequence type) {
      * Ctor.
```
mkordas: @essobedo can you explain here in more detail how using `charset` will impact header?

mkordas:
@essobedo please make link to issue in PR description

src/main/java/org/takes/rs/RsWithBody.java
```diff
@@ -88,7 +89,18 @@ public RsWithBody(final URL url) {
      * @param body Body
      */
     public RsWithBody(final Response res, final CharSequence body) {
-        this(res, body.toString().getBytes(StandardCharsets.UTF_8));
+        this(res, body, StandardCharsets.UTF_8);
+    }
+
+    /**
+     * Ctor.
```
mkordas: @essobedo we need more Javadoc here about the impact of `charset` parameter

mkordas:
@essobedo please see a few comments above

essobedo:
@mkordas Thank you for your remarks, PR updated consequently please review

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -46,25 +47,67 @@
...
      */
-    private static Response make(final Response res, final CharSequence type) {
-        return new RsWithHeader(
-            new RsWithoutHeader(res, RsWithType.HEADER),
-            RsWithType.HEADER, type
-        );
+    private static Response make(final Response res, final CharSequence type,
+        final Charset charset) {
+        final Response response;
+        if (charset == null) {
```
mkordas: @essobedo please do not use `null`

essobedo: @mkordas What do you propose then?

essobedo: @mkordas if you check well I call charset.name() in this method so it is necessary to test if it is null or not, and instead of throwing an IllegalArgumentException, I think that is much better in that case to simply call the code that doesn't add the charset, don't you agree? If not could you please describe what you have in mind?

essobedo:
@mkordas PR updated with more javadoc about the charset as requested

src/main/java/org/takes/rs/RsWithType.java
mkordas: @essobedo `null` in Takes is simply prohibited, please see README. You can try using `Optional` or custom object instead.

essobedo:
@mkordas It is very clear now thank you for the suggestion, here is my new PR

essobedo:
@mkordas new version of the PR with no more null inside as expected so it should be fine now

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -79,11 +137,35 @@ private static Response make(final Response res, final CharSequence type) {
...
+         * @param charset The character set to add in the content type header
+         */
+        public HTML(final Response res, final Charset charset) {
+            this(res, new Opt.Single<Charset>(charset));
+        }
+
+        /**
+         * Constructs a {@code HTML} that will add text/html as the content type
+         * header to the response using the specified charset as charset
+         * parameter value if present..
```
mkordas: @essobedo two dots :)

mkordas:
@essobedo I'm doing final review now, looks nice!

src/test/java/org/takes/rs/RsWithTypeTest.java
```diff
@@ -151,22 +174,38 @@ public void replacesTypeWithHtml() throws Exception {
...
     /**
      * RsWithType.JSON can replace an existing type with application/json.
-     * @throws java.lang.Exception If a problem occurs.
+     * @throws Exception If a problem occurs.
      */
     @Test
     public void replacesTypeWithJson() throws Exception {
         MatcherAssert.assertThat(
             new RsPrint(
+                new RsWithType.JSON(new RsWithType(new RsEmpty(), TYPE_XML))
```
mkordas: @essobedo why we use `TYPE_XML` in JSON test?

src/test/java/org/takes/rs/RsWithTypeTest.java
```diff
@@ -176,22 +215,38 @@ public void replacesTypeWithJson() throws Exception {
...
     /**
      * RsWithType.XML can replace an existing type with text/xml.
-     * @throws java.lang.Exception If a problem occurs.
+     * @throws Exception If a problem occurs.
      */
     @Test
     public void replacesTypeWithXml() throws Exception {
         MatcherAssert.assertThat(
             new RsPrint(
+                new RsWithType.XML(new RsWithType(new RsEmpty(), TYPE_HTML))
```
mkordas: @essobedo similar question here - why HTML in XML test?

src/main/java/org/takes/rs/RsWithBody.java
```diff
@@ -51,7 +52,8 @@
 public final class RsWithBody extends RsWrap {
 
     /**
-     * Ctor.
+     * Constructs a {@code RsWithBody} with the specified body that will be
+     * encoded into UTF-8 by default.
```
mkordas: @essobedo here we state that `UTF-8` is used by default. What about the rest of constructors? Javadocs are not consistent.

mkordas:
@essobedo very nice PR, just couple of additional minor comments from my side and I'm ready to merge

src/main/java/org/takes/rs/RsWithType.java
essobedo: good catch :-)

src/test/java/org/takes/rs/RsWithTypeTest.java
essobedo: Because the goal of this unit test (that was not created by me) is to show that we can replace the previous content type which is text/xml here with a new content type which is application/json in that case.

src/test/java/org/takes/rs/RsWithTypeTest.java
essobedo: same answer then

src/main/java/org/takes/rs/RsWithBody.java
essobedo: @mkordas Unless I missed a constructor but for me actually it is consistent because the other constructors deal with already serialized content such as byte[], InputStream or URL so I have no way to know how it has been encoded.

essobedo:
@mkordas thank you for your remarks, here is a new version of my PR hoping that it is ok this time :-)

src/main/java/org/takes/rs/RsWithBody.java
mkordas: @essobedo makes sense, thanks

src/test/java/org/takes/rs/RsWithTypeTest.java
mkordas: @essobedo right, now I see

src/test/java/org/takes/rs/RsWithTypeTest.java
mkordas: @essobedo thanks

mkordas:
@essobedo nice work, thanks

mkordas:
@rultor merge

yegor256:
@rultor try to merge

mkordas:
@essobedo there are merge conflicts, can you resolve them?

essobedo:
@mkordas I'm on it

essobedo:
@mkordas PR updated

mkordas:
@essobedo thanks!

mkordas:
@rultor merge

essobedo:
@yegor256 Please confirm

essobedo:
@davvd or @yegor256 Please confirm

mkordas:
@essobedo it needs to be @yegor256 - he's an architect here

essobedo:
@yegor256 please confirm

yegor256:
@rultor try to merge

mkordas:
@rultor deploy

davvd:
@ypshenychka please, check this issue for QA compliance, as per [par.24](http://at.teamed.io/policy.html#24)

davvd:
@rultor deploy

ypshenychka:
@davvd Quality is good.

davvd:
> @davvd Quality is good.

@ypshenychka thanks for the review

davvd:
@mkordas 10 mins was added to the account of @ypshenychka (for QA review), in transaction 77294522

Issue #555, opened 2016-01-29
Remove all 'null' usages in code


mkordas:
Takes advertises itself as a library without any `null`. That's not true, I see at least 26 [`null` usages](https://github.com/yegor256/takes/search?l=java&q=null) in Java code.

I'd expect removing `null` from as many places as possible and replacing them with alternatives. 
If some external API actually demands `null` and there's no workaround, we need to admit it in README, and say about few exceptions related to bad external APIs.

I'm fine to solve that issue with PDD approach.

davvd:
@mkordas I am aware of the task, give me some time to find a developer...

davvd:
@mkordas I set milestone here to `1.0`, let me know if it is wrong

davvd:
@mkordas thanks for this report, I added 15 mins to your account, in transaction `56ac901105fa01580100020c`

davvd:
@happy-neko the task is yours,please go ahead

Happy-Neko:
We should make clear what exactly we consider null-free design project-wise. From my point of view:

* Every public API method we expose should not return `null`. Alternatives are throwing exception or returning nullable container if return value is optional (`Optional` in guava / java 8, in takes we have similar `Opt<T>`).
* Do not accept `null` values as method parameters values in out API (up to debate). Alternative approach is to mark parameters with `@NotNull ` / `@Nullable` annotations for self-documentation and static analyser purposes.

Additional consideration:
* We could not stop using 'null' internally because this is how language is designed and core Java library and almost every external library uses it extensively. Workarounds here are pointless.

I have reviewed takes code (except unit tests because they do not expose public API) and found no `null` usages that violates these principles. So it is pretty safe to say that "null-free" is true.

@mkordas what do you think?

mkordas:
@Happy-Neko please see http://www.yegor256.com/2014/05/13/why-null-is-bad.html - it states that null is bad  in general, not only in public API

Happy-Neko:
@mkordas The problem is we can do nothing if foreign API method can return `null` or accepts it as parameter.

mkordas:
@Happy-Neko we can use `Opt` or `Optional` in such cases

Happy-Neko:
> we can use Opt or Optional in such cases

@mkordas We still need to test for `Optional` value presence every time we made `null` check before. 
This is hiding the problem under extra layer of code complexity imho.

@yegor256 What do you think?

yegor256:
[...]

Issue #609, opened 2016-02-12
null value should not be placed in Opt.Single container


Happy-Neko:
`TkMethodsTest` and `TkRetryTest` detected problems with passing `null` value to `Opt.Single` constructor. We should fix these problems and re-enable ignored tests.

davvd:
@yegor256 please do something about it, see [par.21](http://at.teamed.io/policy.html#21)

yegor256:
@Happy-Neko are there puzzles for that problems already in the code?

Happy-Neko:
@yegor256 in PR #610.

yegor256:
@Happy-Neko what exactly is wrong now? this ticket sounds like a task, not a bug. read this: http://www.yegor256.com/2014/11/24/principles-of-bug-tracking.html#5-report-when-it-is-broken

Happy-Neko:
> @yegor256 what exactly is wrong now?

There are NPEs in ignored tests (PR #610).

yegor256:
@Happy-Neko but the tests are ignored. so the build is clean. there are no bugs in the code. you're trying to un-ignore and you can't do it. it's your problem, in your own pull request. we can't blame "master" branch for that.

Issue #608, opened 2016-02-12
Wrap all methods that may return null in Opt.Nullable


Happy-Neko:
All library methods calls that may return `null` should be wrapped in `Opt.Nullable` and checked for value presence with `Opt.has()`.

davvd:
@yegor256 please do something about it, see [par.21](http://at.teamed.io/policy.html#21)

krzyk:
@Happy-Neko This is a todo, that should be code that introduced `Opt.Nullable`, it is not a bug in master branch.

yegor256:
@0crat in

0crat:
> @0crat in ([here](https://github.com/yegor256/takes/issues/608#issuecomment-322126102))

@yegor256 Job `gh:yegor256/takes#608` is in scope.

0crat:
Job `gh:yegor256/takes#608` assigned to @caarlos0, please go ahead ([policy](http://datum.zerocracy.com/pages/policy.html)).

0crat:
@caarlos0 this job was assigned to you 8 days ago.
It will be taken away from you after 10 days from start (this is our policy).

0crat:
Oops! Job `gh:yegor256/takes#608` is not assigned to anyone.

0crat:
Oops! Job `gh:yegor256/takes#608` is not assigned to anyone.

0crat:
Job `gh:yegor256/takes#608` assigned to @neonailol  ([profile](http://www.0crat.com/u/neonailol})). The budget is [fixed](http://datum.zerocracy.com/pages/policy.html#4) and it is 30 minutes. Please, read [§4](http://datum.zerocracy.com/pages/policy.html#4), [§8](http://datum.zerocracy.com/pages/policy.html#8), and [§9](http://datum.zerocracy.com/pages/policy.html#9). If the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html).

neonailol:
@yegor256 Please provide more information about the issue.
For example lets look at `org.takes.rq.RqHrefTest#failsOnAbsentRequestLine`

```java
    @Test(expected = HttpException.class)
    public void failsOnAbsentRequestLine() throws IOException {
        new RqHref.Base(
            new RqSimple(Collections.<String>emptyList(), null)
        ).href();
    }
```

[...]

PR #610, opened 2016-02-12
Opt.Nullable implementation

Happy-Neko:
Fix for #555.

* `Opt.Nullable` container implementation to store possible-null value
* Only non-null values should be placed in`Opt.Single` container


Happy-Neko:
@davvd Please proceed with code review.

davvd:
@happy-neko Thanks, I will find someone to review this PR soon

davvd:
@krzyk review this one, please

src/main/java/org/takes/misc/Opt.java
```diff
@@ -49,6 +51,54 @@
...
+     * @todo #608:30min All library methods calls that may return null
+     *  should be wrapped in Opt.Nullable and checked for value presence
+     *  with Opt.has().
+     */
+    final class Nullable<T> implements Opt<T> {
+        /**
+         * Origin.
+         */
+        private final transient T origin;
+        /**
```
krzyk: @Happy-Neko please add new line between fields

src/main/java/org/takes/misc/Opt.java
```diff
@@ -49,6 +51,54 @@
...
+         */
+        private final transient boolean has;
+
+        /**
+         * Ctor.
+         * @param orgn The possibly-null item to hold
+         */
+        public Nullable(final T orgn) {
+            this.origin = orgn;
+            this.has = orgn != null;
```
krzyk: @Happy-Neko this field is not neccessary, you should check for null in the get method (somehow related http://www.yegor256.com/2016/01/26/defensive-programming.html), you should check for valid values only during retrieval

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Test case for {@link Opt.Nullable}.
+ * @author I. Sokolov (happy.neko@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+public final class NullableTest {
```
krzyk: @Happy-Neko I would rather name it `OptNullableTest`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Item to hold in {@link Opt.Nullable} container.
+     */
+    private transient Object item;
```
krzyk: @Happy-Neko why do you need a field? It couples tests unnecessarily.

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Make item instance.
+     * @throws Exception If something goes wrong
+     */
+    @Before
+    public void setUp() throws Exception {
```
krzyk: @Happy-Neko same here, `setUp` is usually a bad practice

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+        this.item = new Object();
+    }
+
+    /**
+     * Has should return true for non-empty container.
```
krzyk: @Happy-Neko `Opt.Nullable can return true for non-empty container.` (see http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#test-method-names)

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+     * @throws Exception If something goes wrong
+     */
+    @Test
+    public void hasNonNullItemShouldReturnTrue() throws Exception {
```
krzyk: @Happy-Neko `returnsTrueForNonEmptyContainer`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+        MatcherAssert.assertThat(
+            new Opt.Nullable<>(this.item).has(),
+            Matchers.is(true)
+        );
+    }
+
+    /**
+     * Has should return false for empty container.
```
krzyk: @Happy-Neko `Opt.Nullable can return false for empty container.`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+     * @throws Exception If something goes wrong
+     */
+    @Test
+    public void hasNullItemShouldReturnFalse() throws Exception {
```
krzyk: @Happy-Neko `returnsFalseForEmptyContainer`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+        MatcherAssert.assertThat(
+            new Opt.Nullable<>(null).has(),
+            Matchers.is(false)
+        );
+    }
+
+    /**
+     * Get should return item for non-empty container.
```
krzyk: @Happy-Neko `Opt.Nullable can return item for non-empty container.`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+     * @throws Exception If something goes wrong
+     */
+    @Test
+    public void getNonNullItemShouldReturnIt() throws Exception {
```
krzyk: @Happy-Neko `returnsItemForNonEmptyContainer`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+        MatcherAssert.assertThat(
+            new Opt.Nullable<>(this.item).get(),
+            Matchers.sameInstance(this.item)
+        );
+    }
+
+    /**
+     * Get should throw NoSuchElementException for empty container.
```
krzyk: @Happy-Neko `Opt.Nullable can throw NoSuchElementException for empty container`

src/test/java/org/takes/misc/NullableTest.java
```diff
@@ -0,0 +1,98 @@
...
+     * @throws Exception If something goes wrong
+     */
+    @Test(expected = NoSuchElementException.class)
+    public void failsOnGetNullItem() throws Exception {
```
krzyk: @Happy-Neko `throwsNoSuchElementExceptionForEmptyContainer`

src/test/java/org/takes/misc/NullableTest.java
krzyk: @Happy-Neko you could just do `new Object()` inside the tests instead of `this.item`

src/main/java/org/takes/misc/Opt.java
```diff
@@ -66,9 +116,14 @@
...
          * @param orgn Origin
+         * @todo #609:30min null value should not be placed in Opt.Single
+         *  container. TkMethodsTest and TkRetryTest detected problems with
+         *  passing null value to Opt.Single constructor. We should fix these
+         *  problems and re-enable ignored tests.
          */
         public Single(final T orgn) {
-            this.origin = orgn;
+            this.origin =
+                Objects.requireNonNull(orgn, "orgn could not be null");
```
krzyk: @Happy-Neko here I'm wondering, is this class used from outside (initialized by client)? If not then there is no need to check for nulls here, just make sure we don't provide a null to it.

krzyk:
@Happy-Neko please see my comments

Happy-Neko:
@krzyk fixes committed, thanks for the review.

One thing I disagree with is accepting `null` in `Opt.Single` constructor. Since this is container for non-null value we should be defensive here and not instantiate object in invalid state hiding the bug. Adding simple validity check here already revealed at least two bugs in our code (see #609).

krzyk:
@Happy-Neko have a look here http://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html#comment-2014240415

src/test/java/org/takes/misc/OptNullableTest.java
```diff
@@ -0,0 +1,86 @@
...
+    public void returnsItemForNonEmptyContainer() throws Exception {
+        final Object item = new Object();
+        MatcherAssert.assertThat(
+            new Opt.Nullable<>(item).get(),
+            Matchers.sameInstance(item)
+        );
+    }
+
+    /**
+     * {@link Opt.Nullable} can throw NoSuchElementException for empty
```
krzyk: @Happy-Neko link here is not needed

src/test/java/org/takes/misc/OptNullableTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    public void returnsFalseForEmptyContainer() throws Exception {
+        MatcherAssert.assertThat(
+            new Opt.Nullable<>(null).has(),
+            Matchers.is(false)
+        );
+    }
+
+    /**
+     * {@link Opt.Nullable} can return item for non-empty container.
```
krzyk: @Happy-Neko link here is not needed

src/test/java/org/takes/misc/OptNullableTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    public void returnsTrueForNonEmptyContainer() throws Exception {
+        MatcherAssert.assertThat(
+            new Opt.Nullable<>(new Object()).has(),
+            Matchers.is(true)
+        );
+    }
+
+    /**
+     * {@link Opt.Nullable} can return false for empty container.
```
krzyk: @Happy-Neko link here is not needed

src/test/java/org/takes/misc/OptNullableTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Test case for {@link Opt.Nullable}.
+ * @author I. Sokolov (happy.neko@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+public final class OptNullableTest {
+    /**
+     * {@link Opt.Nullable} can return true for non-empty container.
```
krzyk: @Happy-Neko link here is not needed

krzyk:
@Happy-Neko minor thing in the javadocs and we can go

Happy-Neko:
@krzyk javadocs fixed.

krzyk:
@rultor merge pls

yegor256:
@Happy-Neko hey... we're against NULL at all. we don't need to deal with NULL ever. we just don't know what NULL is. by creating such a class we admit that NULL has its place in OOP. but it doesn't

Happy-Neko:
@yegor256 `Nullable` renamed to `Optional`.

yegor256:
@Happy-Neko I don't understand what all this is for. we should **get rid** of NULL everywhere,  not create some wrappers for it. NULL is an anti-pattern, in the first place. it must not exist in Java code. that's what ticket #555 is about.

exper0:
>it must not exist in Java

@yegor256 obviously java is created with other thoughs. `null` considered ok by language creators, it is used in every single library in any hava project(and in JDK itself). Creators of Java just removed everything what they don't like from C++ and got actually Java. Don't you think of creating your own programming language?

yegor256:
@exper0 Java creators made many mistakes. we should not just blindly follow their ideas. NULL is a bad thing, no matter who created it or where.

exper0:
@yegor256 just curious - do you know any language without null or nil concept? I understand your position - in perfect world null doesn't make sense. However, java HashMap may return null, there are lot of such examples. Does it mean we shouldn't use the code where null is normal value?

yegor256:
@exper0 I don't know any languages without NULL. I don't have an answer about JDK classes. let's try to think what to do with them. ideally, we should stay away from them. but it's not always possible. ideas?

Happy-Neko:
@yegor256 
> not create some wrappers for it. NULL

`Optional<T>` is not wrapper for `null`. It is container for specific type that may be empty. This is how possible absence of value is represented in languages that forbid `null` references (for ex. Swift). We already have `Opt.Single` for empty containers. Is it wrapper for `null`, what do you think?

Anyway we could not throw away API methods that may return `null` whether we like it or not. And this means testing return values for nullability. If you want to reduce 'null' literal usage in the code this is the way:

```java
final Optional<Type> result = someUglyMethod();
if (result.has()) {
    // method returned something; extract with result.get()
} else {
    // method returned nothing
}
```

yegor256:
@Happy-Neko yes, this is probably the only option we have... let's go for it

Happy-Neko:
@yegor256 done.

src/main/java/org/takes/misc/Optional.java
```diff
@@ -0,0 +1,85 @@
...
+        return new Optional<>(null);
+    }
+
+    /**
+     * Returns the contained item.
+     * @return Item instance
+     */
+    public T get() {
+        if (this.origin == null) {
+            throw new NoSuchElementException(
```
yegor256: @Happy-Neko see http://www.yegor256.com/2015/01/21/if-then-throw-else.html

src/main/java/org/takes/misc/Optional.java
```diff
@@ -0,0 +1,85 @@
...
+    public Optional(final T orgn) {
+        this.origin = orgn;
+    }
+
+    /**
+     * Returns empty container.
+     * @param <T> Type of item
+     * @return Empty {@link Optional}
+     */
+    public static <T> Optional<T> empty() {
```
yegor256: @Happy-Neko static methods are evil. no exceptions. see http://www.yegor256.com/2014/05/05/oop-alternative-to-utility-classes.html

yegor256:
@Happy-Neko first, I didn't understand why did you rename Opt to Optional. second, your pull request doesn't prevent us from using null again. There must be some check for that. @krzyk please, take a look

krzyk:
@Happy-Neko OK, after looking at the places where null is used in the code I think you approached in to general way, while actually you should go and remove all the nulls possible.
Here is a good start:
* BkSafe, assert not needed, workaround for poor exception handling
* CcBase64 - all asserts are not needed if qulice is upgraded to 0.16.1
* FkHitRefresh - this null check is not needed if earliler we check that the directory exists
* FtBasic - assert not needed, workaround for poor exception handling
* Href - null check not needed, one has to check if key exists instead
* Items - check if File is directory first, then null should not happen
* MainRemote - not sure here, probably more investigation needed
* Options - check if key exists instead of null check, and in the matcher case, check matcher.groupCount
* PsBasic - check if key exists (containsKey)
* PsGoogle - check if key exists (Json is a Map)
* RqCookies - again containsKey

and so on, each case has to be analyzed, and I think in majority of the cases null can be removed, so there is no point in introducing another class that will have ` != null`.

Happy-Neko:
@yegor256 fixes committed.

Happy-Neko:
@yegor256 @krzyk 
What work need to be done to have this PR merged? Please keep in mind this ticket allocated time budget 30 min.

krzyk:
@Happy-Neko You can use todos for some parts of the work. You didn't address @yegor256 comments (about renaming to Optional, and about a check to prevent using nulls).
And I still think the correct approach would be to just remove the nulls from the code by using a correct API (e.g. map.get(..) != null is not correct, one should use map.containsKey()) and it is quite easy, not controversial. You could leave some of those nulls in a todo, together with a check that would fail a build if a `null` appears in the code.

This would be much easier then what you are doing in current PR, where you assume that we can't get rid of JDKs nulls, it might not be possible, but you didn't try it.

Happy-Neko:
@krzyk @yegor256 

1. `Optional` is different class (not implementing `Opt` interface). `Opt` was removed as redundant because `Optional` can serve both as empty and non-empty container.

2. Checking for `null` return values or changing API methods is outside of this PR's scope  - implementing container type.

3. Decision to wrap return values in Optional was confirmed by @yegor256 here https://github.com/yegor256/takes/pull/610#issuecomment-190291755 

I am sticking with it due to task's estimated time budget. 

4. TODO is already created #608.

src/main/java/org/takes/facets/auth/PsAll.java
```diff
@@ -60,12 +60,13 @@ public PsAll(final List<? extends Pass> passes, final int identity) {
...
-    public final Opt<Identity> enter(final Request request) throws IOException {
-        final Opt<Identity> result;
+    public final Optional<Identity> enter(final Request request)
+        throws IOException {
+        final Optional<Identity> result;
         if (this.allMatch(request)) {
             result = this.all.get(this.index).enter(request);
         } else {
-            result = new Opt.Empty<Identity>();
+            result = new Optional<>(null);
```
krzyk: @Happy-Neko why change a perfectly fine `Opt.Empty` to a optional that is called with a `null`? This is ugly.

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -159,16 +159,16 @@ public Fake(final boolean cond) {
...
             if (this.condition) {
-                user = new Opt.Single<Identity>(
+                user = new Optional<Identity>(
                     new Identity.Simple(
                         String.format("urn:basic:%s", usr)
                     )
                 );
             } else {
-                user = new Opt.Empty<>();
+                user = new Optional<>(null);
```
krzyk: @Happy-Neko same here, why create something with a null in constructor?

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -183,8 +183,8 @@ public Fake(final boolean cond) {
      */
     public static final class Empty implements PsBasic.Entry {
         @Override
-        public Opt<Identity> enter(final String user, final String pwd) {
-            return new Opt.Empty<>();
+        public Optional<Identity> enter(final String user, final String pwd) {
+            return new Optional<>(null);
```
krzyk: @Happy-Neko and here, should be `Opt.Empty`

src/main/java/org/takes/facets/auth/PsBasic.java
```diff
@@ -254,7 +254,7 @@ public Default(final String... users) {
                     throw new IllegalStateException(ex);
                 }
             } else {
-                identity = new Opt.Empty<>();
+                identity = new Optional<>(null);
```
krzyk: @Happy-Neko should be `Opt.Empty` (please see https://en.wikipedia.org/wiki/Null_Object_pattern)

src/main/java/org/takes/facets/auth/PsByFlag.java
```diff
@@ -96,10 +96,10 @@ public PsByFlag(final String flg, final Map<Pattern, Pass> map) {
     }
 
     @Override
-    public Opt<Identity> enter(final Request req) throws IOException {
+    public Optional<Identity> enter(final Request req) throws IOException {
         final Iterator<String> flg = new RqHref.Base(req).href()
             .param(this.flag).iterator();
-        Opt<Identity> user = new Opt.Empty<Identity>();
+        Optional<Identity> user = new Optional<>(null);
```
krzyk: @Happy-Neko as in othere cases, should be `Opt.Empty` (or `Optional.Empty`)

src/main/java/org/takes/facets/auth/PsChain.java
```diff
@@ -64,8 +64,8 @@ public PsChain(final Iterable<Pass> list) {
     }
 
     @Override
-    public Opt<Identity> enter(final Request req) throws IOException {
-        Opt<Identity> user = new Opt.Empty<Identity>();
+    public Optional<Identity> enter(final Request req) throws IOException {
+        Optional<Identity> user = new Optional<>(null);
```
krzyk: @Happy-Neko as in othere cases, should be `Opt.Empty` (or `Optional.Empty`)

src/main/java/org/takes/facets/auth/PsCookie.java
```diff
@@ -96,16 +96,18 @@ public PsCookie(final Codec cdc, final String name, final long days) {
...
+        final Optional<Identity> user;
         if (cookies.hasNext()) {
-            user = new Opt.Single<Identity>(
+            user = new Optional<>(
                 this.codec.decode(
                     cookies.next().getBytes(StandardCharsets.UTF_8)
                 )
             );
+        } else {
+            user = new Optional<>(null);
```
krzyk: @Happy-Neko as in othere cases, should be `Opt.Empty` (or `Optional.Empty`)

src/main/java/org/takes/facets/auth/PsEmpty.java
```diff
@@ -41,8 +41,8 @@
 public final class PsEmpty implements Pass {
 
     @Override
-    public Opt<Identity> enter(final Request request) {
-        return new Opt.Empty<Identity>();
+    public Optional<Identity> enter(final Request request) {
+        return new Optional<>(null);
```
krzyk: @Happy-Neko as in othere cases, should be `Opt.Empty` (or `Optional.Empty`)

src/main/java/org/takes/facets/auth/PsFake.java
```diff
@@ -54,14 +54,14 @@ public PsFake(final boolean cond) {
...
+    public Optional<Identity> enter(final Request request) {
+        final Optional<Identity> user;
         if (this.condition) {
-            user = new Opt.Single<Identity>(
+            user = new Optional<Identity>(
                 new Identity.Simple("urn:test:1")
             );
         } else {
-            user = new Opt.Empty<Identity>();
+            user = new Optional<>(null);
```
krzyk: @Happy-Neko as in othere cases, should be `Opt.Empty` (or `Optional.Empty`)

src/main/java/org/takes/facets/fallback/FbChain.java
```diff
@@ -64,11 +64,11 @@ public FbChain(final Iterable<Fallback> fallbacks) {
         super(
             new Fallback() {
                 @Override
-                public Opt<Response> route(final RqFallback req)
+                public Optional<Response> route(final RqFallback req)
                     throws IOException {
-                    Opt<Response> rsp = new Opt.Empty<Response>();
+                    Optional<Response> rsp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fallback/FbEmpty.java
```diff
@@ -46,8 +46,8 @@ public FbEmpty() {
         super(
             new Fallback() {
                 @Override
-                public Opt<Response> route(final RqFallback req) {
-                    return new Opt.Empty<Response>();
+                public Optional<Response> route(final RqFallback req) {
+                    return new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fallback/FbLog4j.java
```diff
@@ -50,10 +50,10 @@ public FbLog4j() {
         super(
             new Fallback() {
                 @Override
-                public Opt<Response> route(final RqFallback req)
+                public Optional<Response> route(final RqFallback req)
                     throws IOException {
                     FbLog4j.log(req);
-                    return new Opt.Empty<Response>();
+                    return new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fallback/FbSlf4j.java
```diff
@@ -59,10 +59,10 @@ public FbSlf4j() {
         super(
             new Fallback() {
                 @Override
-                public Opt<Response> route(final RqFallback req)
+                public Optional<Response> route(final RqFallback req)
                     throws IOException {
                     FbSlf4j.log(req);
-                    return new Opt.Empty<Response>();
+                    return new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fallback/FbStatus.java
```diff
@@ -154,11 +154,13 @@ public FbStatus(final Condition<Integer> check, final Fallback fallback) {
...
                 @Override
-                public Opt<Response> route(final RqFallback req)
+                public Optional<Response> route(final RqFallback req)
                     throws IOException {
-                    Opt<Response> rsp = new Opt.Empty<Response>();
+                    final Optional<Response> rsp;
                     if (check.fits(req.code())) {
                         rsp = fallback.route(req);
+                    } else {
+                        rsp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkAnonymous.java
```diff
@@ -73,13 +73,13 @@ public FkAnonymous(final Take tke) {
...
+    public Optional<Response> route(final Request req) throws IOException {
         final Identity identity = new RqAuth(req).identity();
-        final Opt<Response> resp;
+        final Optional<Response> resp;
         if (identity.equals(Identity.ANONYMOUS)) {
-            resp = new Opt.Single<Response>(this.take.act(req));
+            resp = new Optional<>(this.take.act(req));
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkChain.java
```diff
@@ -69,10 +69,10 @@ public FkChain(final Collection<Fork> forks) {
     }
 
     @Override
-    public Opt<Response> route(final Request request) throws IOException {
-        Opt<Response> response = new Opt.Empty<Response>();
+    public Optional<Response> route(final Request request) throws IOException {
+        Optional<Response> response = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkContentType.java
```diff
@@ -64,12 +64,12 @@ public FkContentType(final String atype, final Response response) {
...
-    public Opt<Response> route(final Request req) throws IOException {
-        final Opt<Response> resp;
+    public Optional<Response> route(final Request req) throws IOException {
+        final Optional<Response> resp;
         if (FkContentType.getType(req).contains(this.type)) {
-            resp = new Opt.Single<Response>(this.origin);
+            resp = new Optional<>(this.origin);
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkEncoding.java
```diff
@@ -101,12 +101,12 @@ public FkEncoding(final String enc, final Response response) {
                 )
             );
             if (items.contains(this.encoding)) {
-                resp = new Opt.Single<Response>(this.origin);
+                resp = new Optional<>(this.origin);
             } else {
-                resp = new Opt.Empty<Response>();
+                resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkEncoding.java
```diff
@@ -101,12 +101,12 @@ public FkEncoding(final String enc, final Response response) {
...
             }
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkHitRefresh.java
```diff
@@ -123,18 +123,18 @@ public FkHitRefresh(final File file, final Runnable cmd,
...
         if (header.hasNext()) {
             if (this.expired()) {
                 this.exec.run();
                 FkHitRefresh.touch(this.last);
             }
-            resp = new Opt.Single<Response>(this.take.act(req));
+            resp = new Optional<>(this.take.act(req));
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkHost.java
```diff
@@ -71,16 +71,17 @@ public FkHost(final String host, final Take take) {
...
                 ).single("host");
-                final Opt<Response> rsp;
+                final Optional<Response> rsp;
                 if (host.toLowerCase(Locale.ENGLISH)
                     .equals(hst.toLowerCase(Locale.ENGLISH))) {
-                    rsp = new Opt.Single<>(take.act(req));
+                    rsp = new Optional<>(take.act(req));
                 } else {
-                    rsp = new Opt.Empty<>();
+                    rsp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkMethods.java
```diff
@@ -96,13 +96,13 @@ public FkMethods(final Collection<String> mtds, final Take tke) {
...
+    public Optional<Response> route(final Request req) throws IOException {
         final String mtd = new RqMethod.Base(req).method();
-        final Opt<Response> resp;
+        final Optional<Response> resp;
         if (this.methods.contains(mtd)) {
-            resp = new Opt.Single<Response>(this.take.act(req));
+            resp = new Optional<>(this.take.act(req));
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkParams.java
```diff
@@ -84,15 +84,15 @@ public FkParams(final String param, final Pattern ptn, final Take tke) {
...
             .param(this.name).iterator();
-        final Opt<Response> resp;
+        final Optional<Response> resp;
         if (params.hasNext()
             && this.pattern.matcher(params.next()).matches()) {
-            resp = new Opt.Single<Response>(this.take.act(req));
+            resp = new Optional<>(this.take.act(req));
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkRegex.java
```diff
@@ -205,7 +205,7 @@ public InputStream body() throws IOException {
                 )
             );
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/facets/fork/FkTypes.java
```diff
@@ -64,12 +64,12 @@ public FkTypes(final String list, final Response response) {
...
-    public Opt<Response> route(final Request req) throws IOException {
-        final Opt<Response> resp;
+    public Optional<Response> route(final Request req) throws IOException {
+        final Optional<Response> resp;
         if (FkTypes.accepted(req).contains(this.types)) {
-            resp = new Opt.Single<Response>(this.origin);
+            resp = new Optional<>(this.origin);
         } else {
-            resp = new Opt.Empty<Response>();
+            resp = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -76,7 +76,7 @@ private static Request parse(final InputStream input) throws IOException {
         boolean eof = true;
         final Collection<String> head = new LinkedList<String>();
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        Opt<Integer> data = new Opt.Empty<Integer>();
+        Optional<Integer> data = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/misc/Opt.java
```diff
@@ -1,104 +0,0 @@
-/**
```
krzyk: @Happy-Neko why did you remove this class? It was much cleaner than the `Optional` you added

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -99,15 +99,15 @@ private static Request parse(final InputStream input) throws IOException {
...
+                data = new Optional<>(input.read());
+                final Optional<String> header = newHeader(data, baos);
                 if (header.has()) {
                     head.add(header.get());
                 }
                 continue;
             }
             baos.write(legalCharacter(data, baos, head.size() + 1));
-            data = new Opt.Empty<Integer>();
+            data = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -130,11 +130,11 @@ public InputStream body() {
      * @param baos Current read header
      * @return Read header
      */
-    private static Opt<String> newHeader(final Opt<Integer> data,
+    private static Optional<String> newHeader(final Optional<Integer> data,
         final ByteArrayOutputStream baos) {
-        Opt<String> header = new Opt.Empty<String>();
+        Optional<String> header = new Optional<>(null);
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -60,7 +60,7 @@
      * @param type Content type
      */
     public RsWithType(final Response res, final CharSequence type) {
-        this(res, type, new Opt.Empty<Charset>());
+        this(res, type, new Optional<Charset>(null));
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -142,7 +142,7 @@ private static Response make(final Response res, final CharSequence type,
          * @param res Original response
          */
         public HTML(final Response res) {
-            this(res, new Opt.Empty<Charset>());
+            this(res, new Optional<Charset>(null));
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -187,7 +187,7 @@ private HTML(final Response res, final Opt<Charset> charset) {
          * @param res Original response
          */
         public JSON(final Response res) {
-            this(res, new Opt.Empty<Charset>());
+            this(res, new Optional<Charset>(null));
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

src/main/java/org/takes/rs/RsWithType.java
```diff
@@ -277,7 +277,7 @@ private XML(final Response res, final Opt<Charset> charset) {
          * @param res Original response
          */
         public Text(final Response res) {
-            this(res, new Opt.Empty<Charset>());
+            this(res, new Optional<Charset>(null));
```
krzyk: @Happy-Neko again, `Opt.Empty` is cleaner

krzyk:
@Happy-Neko @yegor256 do we proceed with this PR? It is already 1.5 month old.

krzyk:
@Happy-Neko I don't understand why you removed `Opt.Empty` and replaced with `Optional`, `Opt.Empty` should stay as it was, and all of its usages should be unchanged. What you did introduced additional nulls to the project which is in contradiction what this ticket is for.
As @yegor256 confirmed that we need a container for nulls, create just that, don't throw away existing `Opt` classes.
Your new null container class will be used (probably in future tasks - add todo) to replace places where we have null and can't get rid of them.

krzyk:
@yegor256 is Happy-Neko still with us?

Happy-Neko:
@davvd @yegor256 @krzyk 
I could not work on assigned issues, please reassign them. 
Also netbout's google login is not working and I could not set "away" status.

krzyk:
@Happy-Neko so, please close this PR if you can't work on it now

krzyk:
@yegor256 @Happy-Neko can we drop this PR? It doesn't seem like it will finish.

Happy-Neko:
@krzyk ok.

davvd:
@ypshenychka review this ticket please, for compliance with our [quality rules](http://at.teamed.io/qa.html)

ypshenychka:
@davvd Quality is good.

davvd:
> @davvd Quality is good.

@ypshenychka many thanks

davvd:
@krzyk just added 10 mins to @ypshenychka (for QA), payment ID is `84796490`; Much appreciated! **1 hour and 18 mins** was added to your account, payment ID `84796509`, time spent is 1730 hours and 33 mins; you're getting extra minutes here (c=63); added +78 to your rating, now it is equal to [+3761](http://www.netbout.com/b/35035?open=rating)

Issue #495, opened 2016-01-03
IOExceptions with bootstrap


piotrkot:
Here is a sample project with 2 IOExceptions being reproduced.
https://github.com/piotrkot/takesbug

yegor256:
@piotrkot hm... interesting. thanks for reporting!

davvd:
@piotrkot this task will get someone's attention soon

davvd:
@piotrkot since there is no milestone yet I set it to "1.0"

davvd:
@rajithd can you please help? Keep in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). If you have any technical questions, don't hesitate to ask right here

The cost of this task is **30 mins** (this is exactly how much will be paid, not less not more), when the task is [done](http://www.xdsd.org/2014/04/15/definition-of-done.html)

davvd:
@rajithd check this ["no obligations principle"](http://www.xdsd.org/2014/04/13/no-obligations-principle.html).. This task is on your name for at least 15 days. If you can't close it within the next 48 hours I will have to assign someone else to it. [This article](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html) should help if you're stuck

-30 added to your rating, current score is: [-60](http://www.netbout.com/b/36278?open=rating)

davvd:
@rajithd the task is overdue, and I have to re-assign it to someone else. Please, stop working with it immediately. In general, we're against overdue tasks, check [this page](http://www.xdsd.org/2014/04/13/no-obligations-principle.html)... -60 added to your rating, current score is: [-180](http://www.netbout.com/b/36278?open=rating)

davvd:
@essobedo the task is yours please proceed

essobedo:
@piotrkot As workaround you can use _Bug.class.getResource_ instead of _Bug.class.getResourceAsStream_.

essobedo:
@davvd here is my [PR](https://github.com/yegor256/takes/pull/630), please find a reviewer, thanks in advance.

davvd:
> @davvd here is my [PR](https://github.com/yegor256/takes/pull/630), please find a reviewer, thanks in advance.

@essobedo OK

piotrkot:
@essobedo Yes, I know it could work without the stream. IMHO the constructor with inputstream is a mistake.

essobedo:
[...]

PR #630, opened 2016-02-29
495: Store the content of the stream into a temporary file for reuse

essobedo:
Fix for https://github.com/yegor256/takes/issues/495

this issue is related to the fact that the code in the use case relies on [this constructor](https://github.com/yegor256/takes/blob/31c1ad15ce4b2ed02b616e5f2fdce10033411ebe/src/main/java/org/takes/rs/RsWithBody.java#L155) to build its response so when the InputStream has been consumed once, we get an IOException on the next calls.

The idea of this PR is to store the content of the input stream into a temporary file in order to be able to read the content as many times as we want and be able to do it securely from concurrent threads which was not the case otherwise as most InputStream implementations have not been designed to be shared which could affect the thread safety of the whole class RsWithBody.


davvd:
@essobedo Thanks, I will find someone to review this PR soon

davvd:
@krzyk it's yours,please review

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.URL;
+
+/**
+ * The body of a response used by {@link RsWithBody}.
+ *
+ * @author Nicolas Filotto (nicolas.filotto@gmail.com)
+ * @version $Id$
+ * @since 0.31
```
krzyk: @essobedo since 0.32

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+
+        @Override
+        public InputStream input() throws IOException {
+            return this.url.openStream();
+        }
+
+        @Override
+        public int length() throws IOException {
+            try (final InputStream input = this.url.openStream()) {
+                return input.available();
```
krzyk: @essobedo `available` is not reliable (and optionally overriden), are you sure that the one from URL will give the whole file size?

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+         * The content of the body in a byte array.
+         */
+        private final transient byte[] bytes;
+
+        /**
+         * Constructs an {@code ByteArrayContent} with the specified byte array.
+         * @param content The content of the body.
+         */
+        ByteArrayContent(final byte[] content) {
+            //@checkstyle ArrayIsStoredDirectly (1 line)
```
krzyk: @essobedo you should clone the array and store it

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+         */
+        private static File content(final InputStream input) {
+            final File file;
+            try {
+                file = File.createTempFile("InputStreamContent", "tmp");
+                file.deleteOnExit();
+                //@checkstyle MagicNumberCheck (1 line)
+                final byte[] buffer = new byte[4096];
+                //@checkstyle CascadeIndentationCheck (2 lines)
+                try (final InputStream content = input;
```
krzyk: @essobedo you can use `Files.copy` to copy InputStream data into file/path

src/main/java/org/takes/rs/BodyContent.java
essobedo: @krzyk Actually it is an existing code that I moved into this class as you can see [here](https://github.com/yegor256/takes/blob/31c1ad15ce4b2ed02b616e5f2fdce10033411ebe/src/main/java/org/takes/rs/RsWithBody.java#L120). The only thing that I added is the try-with-resources in order to close the stream which was not done before.

src/test/java/org/takes/http/FtBasicTest.java
```diff
@@ -196,4 +204,33 @@ public void exec(final URI home) throws IOException {
         );
     }
 
+    /**
+     * FtBasic can consume twice the input stream.
+     * @throws IOException If some problem inside
+     */
+    @Test
+    public void consumesTwiceInputStream() throws IOException {
```
krzyk: @essobedo this test is missing asserts or mockito.verify at least

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+     * @return The length of the stream.
+     * @throws IOException in case the length of the stream could not be
+     *  retrieved.
+     */
+    int length() throws IOException;
+
+    /**
+     * Content of a body based on an {@link URL}.
+     */
+    final class URLContent implements BodyContent {
```
krzyk: @essobedo plese create a test for this class

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+            try (final InputStream input = this.url.openStream()) {
+                return input.available();
+            }
+        }
+    }
+
+    /**
+     * Content of a body based on a byte array.
+     */
+    final class ByteArrayContent implements BodyContent {
```
krzyk: @essobedo plese create a test for this class

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,182 @@
...
+        @Override
+        public int length() {
+            return this.bytes.length;
+        }
+    }
+
+    /**
+     * The content of the body based on an {@link InputStream}.
+     */
+    final class InputStreamContent implements BodyContent {
```
krzyk: @essobedo plese create a test for this class

krzyk:
@essobedo please see my comments

src/test/java/org/takes/http/FtBasicTest.java
essobedo: @krzyk What about assertStatus and assertBody that you have at each call? it is not enough?

src/main/java/org/takes/rs/BodyContent.java
essobedo: @krzyk good one! I forgot it

essobedo:
@krzyk thx for the remarks, here is a new version of the PR

src/main/java/org/takes/rs/BodyContent.java
krzyk: @essobedo OK

src/test/java/org/takes/http/FtBasicTest.java
krzyk: @essobedo sorry, you are right, I didn't notice them

src/test/java/org/takes/rs/ByteArrayContentTest.java
```diff
@@ -0,0 +1,60 @@
...
+ * Test case for {@link BodyContent.ByteArrayContent}.
+ *
+ * @author Nicolas Filotto (nicolas.filotto@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+public final class ByteArrayContentTest {
+
+    /**
+     * BodyContent.ByteArrayContentTest can work.
```
krzyk: @essobedo The name in FtBasicTest is bad, don't reuse it, explain what the tests does (and remember about http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#test-method-names)

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,62 @@
...
+ * Test case for {@link BodyContent.InputStreamContent}.
+ *
+ * @author Nicolas Filotto (nicolas.filotto@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+public final class InputStreamContentTest {
+
+    /**
+     * BodyContent.InputStreamContentTest can work.
```
krzyk: @essobedo don't copy bad descriptions

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,96 @@
...
+     * Clean the test.
+     * @throws IOException In case the temporary file could not be deleted.
+     */
+    @After
+    public void clean() throws IOException {
+        Files.deleteIfExists(this.file);
+    }
+
+    /**
+     * BodyContent.URLContent can work.
```
krzyk: @essobedo don't copy bad descriptions

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,96 @@
...
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void justWorks() throws Exception {
```
krzyk: @essobedo don't copy bad test method names

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,62 @@
...
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void justWorks() throws Exception {
```
krzyk: @essobedo don't copy bad test method names

src/test/java/org/takes/rs/ByteArrayContentTest.java
```diff
@@ -0,0 +1,60 @@
...
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void justWorks() throws Exception {
```
krzyk: @essobedo don't copy bad test method names

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * @author Nicolas Filotto (nicolas.filotto@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+public final class URLContentTest {
+
+    /**
+     * The temporary file.
+     */
+    private transient Path file;
```
krzyk: @essobedo using fields in tests is a bad practice, all the data needed for the test should be in the test method

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Initialize the test.
+     * @throws IOException In case the temporary file could not be created.
+     */
+    @Before
+    public void init() throws IOException {
+        this.file = Files.createTempFile("URLContentTest", "tmp");
```
krzyk: @essobedo move it to the test method

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Clean the test.
+     * @throws IOException In case the temporary file could not be deleted.
+     */
+    @After
+    public void clean() throws IOException {
+        Files.deleteIfExists(this.file);
```
krzyk: @essobedo move it to the test method

src/test/java/org/takes/rs/URLContentTest.java
krzyk: @essobedo the only exception is for `@Rule` and `@Parameters`, so here you could use `@Rule` with `TemporaryFolder` instead of using Files.createTempFile(), you wouldn't need to register the delete on exit in such case

krzyk:
@essobedo please see my additional comments

src/test/java/org/takes/rs/ByteArrayContentTest.java
essobedo: @krzyk I simply did a copy/paste from FtBasicTest.justWorks written by Yegor himself so I assumed that it was correct.

krzyk: @essobedo sometimes there is a rush, but we should avoid such descriptions/method names

essobedo:
@krzyk Here is the new version of the PR, thanks for the remarks

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public final class InputStreamContentTest {
+
+    /**
+     * BodyContent.InputStreamContentTest can provide the expected input.
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void returnsCorrectInput() throws Exception {
+        final String result = "Hello returnsCorrectInput!";
```
krzyk: @essobedo inline this variable

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+        final byte[] bytes = result.getBytes(StandardCharsets.UTF_8);
+        final ByteArrayInputStream input = new ByteArrayInputStream(bytes);
```
krzyk: @essobedo inline this variable

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+        final BodyContent.InputStreamContent content =
```
krzyk: @essobedo and this one as well can be inlined, and let's see if it is readable

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * BodyContent.InputStreamContentTest can provide the expected length.
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void returnsCorrectLength() throws Exception {
+        final String result = "Hello returnsCorrectLength!";
+        final byte[] bytes = result.getBytes(StandardCharsets.UTF_8);
```
krzyk: @essobedo inline this variable (see http://www.yegor256.com/2014/04/27/typical-mistakes-in-java-code.html#one-time-variables)

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+        final ByteArrayInputStream input = new ByteArrayInputStream(bytes);
```
krzyk: @essobedo and this one can also be inlined

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+        final BodyContent.InputStreamContent content =
```
krzyk: @essobedo and this also

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * BodyContent.URLContent can provide the expected input.
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void returnsCorrectInput() throws Exception {
+        final Path file = URLContentTest.createTempFile();
+        try {
+            final String result = "Hello returnsCorrectInput!";
```
krzyk: @essobedo this variable can be inlined

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * BodyContent.URLContent can provide the expected length.
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void returnsCorrectLength() throws Exception {
+        final Path file = URLContentTest.createTempFile();
+        try {
+            final String result = "Hello returnsCorrectLength!";
+            final byte[] bytes = result.getBytes(StandardCharsets.UTF_8);
```
krzyk: @essobedo this variable can be inlined

src/test/java/org/takes/rs/URLContentTest.java
```diff
@@ -0,0 +1,105 @@
...
+            try (final InputStream input = new ByteArrayInputStream(bytes)) {
+                Files.copy(input, file, StandardCopyOption.REPLACE_EXISTING);
+            }
+            final BodyContent.URLContent content =
```
krzyk: @essobedo this variable can be inlined

src/test/java/org/takes/rs/ByteArrayContentTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * BodyContent.ByteArrayContentTest can provide the expected length.
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void returnsCorrectLength() throws Exception {
+        final String result = "Hello returnsCorrectLength!";
+        final byte[] bytes = result.getBytes(StandardCharsets.UTF_8);
```
krzyk: @essobedo inline this

src/test/java/org/takes/rs/ByteArrayContentTest.java
```diff
@@ -0,0 +1,72 @@
...
+        final BodyContent.ByteArrayContent content =
```
krzyk: @essobedo inline this

src/test/java/org/takes/rs/ByteArrayContentTest.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public final class ByteArrayContentTest {
+
+    /**
+     * BodyContent.ByteArrayContentTest can provide the expected input.
+     * @throws Exception If some problem inside.
+     */
+    @Test
+    public void returnsCorrectInput() throws Exception {
+        final String result = "Hello returnsCorrectInput!";
```
krzyk: @essobedo inline this

src/test/java/org/takes/rs/ByteArrayContentTest.java
```diff
@@ -0,0 +1,72 @@
...
+        final byte[] bytes = result.getBytes(StandardCharsets.UTF_8);
+        final BodyContent.ByteArrayContent content =
```
krzyk: @essobedo inline this

src/test/java/org/takes/http/FtBasicTest.java
```diff
@@ -196,4 +204,33 @@ public void exec(final URI home) throws IOException {
...
+        final String result = "Hello World!";
+        final ByteArrayInputStream input =
```
krzyk: @essobedo inline this

krzyk:
@essobedo few variables have to inlined and we are ready to go

essobedo:
@krzyk please check this new version

krzyk:
@essobedo you missed one comnent

essobedo:
@krzyk my bad, here is the new version

krzyk:
@rultor merge pls

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,176 @@
...
+import java.nio.file.StandardCopyOption;
+
+/**
+ * The body of a response used by {@link RsWithBody}.
+ *
+ * @author Nicolas Filotto (nicolas.filotto@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+interface BodyContent {
```
yegor256: @essobedo why not just `Body`?

src/main/java/org/takes/rs/BodyContent.java
```diff
@@ -0,0 +1,176 @@
...
+            return this.bytes.length;
+        }
+    }
+
+    /**
+     * The content of the body based on an {@link InputStream}.
+     */
+    final class InputStreamContent implements BodyContent {
+
+        /**
```
yegor256: @essobedo I would do it like this: `class Stream implements Body`

src/test/java/org/takes/rs/InputStreamContentTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.Test;
+
+/**
+ * Test case for {@link BodyContent.InputStreamContent}.
+ *
+ * @author Nicolas Filotto (nicolas.filotto@gmail.com)
+ * @version $Id$
+ * @since 0.32
+ */
+public final class InputStreamContentTest {
```
yegor256: @essobedo all these tests should stay in `BodyTest` class. the rule is that one `.java` file should match exactly one `Test.java` file

yegor256:
@essobedo see my comments

essobedo:
@yegor256 thx for the remarks, here is the new version of the PR, please check

essobedo:
@yegor256 Could you please recheck, thank you in advance.

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,176 @@
...
+            try (final InputStream input = this.url.openStream()) {
+                return input.available();
+            }
+        }
+    }
+
+    /**
+     * Content of a body based on a byte array.
+     */
+    final class ByteArrayContent implements Body {
```
yegor256: @essobedo same here. `Body.ByteArray` looks much nicer to me than `Body.ByteArrayContent`. right?

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,176 @@
...
+     * @return The length of the stream.
+     * @throws IOException in case the length of the stream could not be
+     *  retrieved.
+     */
+    int length() throws IOException;
+
+    /**
+     * Content of a body based on an {@link URL}.
+     */
+    final class URLContent implements Body {
```
yegor256: @essobedo same here `Body.URL` is good enough

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,176 @@
...
+         * Gives the {@code File} that contains the content of the provided
+         * {@link InputStream}.
+         * @param input The stream to store into the file.
+         * @return The {@code File} in which we stored the content of the
+         *  stream.
+         */
+        private static File content(final InputStream input) {
+            final File file;
+            try {
+                file = File.createTempFile("InputStreamContent", "tmp");
```
yegor256: @essobedo it's better to do `this.getClass().getName()`

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,176 @@
...
+                file.deleteOnExit();
```
yegor256: @essobedo that's a tricky part... this means that the file will be deleted with JVM stops. this may happen in a year. the disc will be full of temp files by then...

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,176 @@
...
+                try (final InputStream content = input) {
+                    Files.copy(
+                        content,
+                        Paths.get(file.getAbsolutePath()),
+                        StandardCopyOption.REPLACE_EXISTING
+                    );
+                }
+            } catch (final IOException ex) {
```
yegor256: @essobedo I would rather let it float up...

yegor256:
@essobedo a few more comments. catching an exception is not a good idea in general: http://www.yegor256.com/2015/07/09/catch-if-cant-otherwise.html

src/main/java/org/takes/rs/Body.java
essobedo: ok as you like

src/main/java/org/takes/rs/Body.java
essobedo: ok

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 ok I will use _Body.Stream.class.getName()_ as it is a static method here

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 If I do so I have 2 options: 

1. I can throw this exception in the constructor which will mean modifying several classes in the current code base to throw the same exception everywhere which could be seen by the end-user as a broken API.
2. I can lazily create the file when calling input() or length() as they both throw this king of Exception but it would mean that my class is no more immutable which is not what you want I guess.

For me both solutions were not acceptable that is why I did that but if you disagree, please tell me, I will do whatever you prefer.

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 I added a comment to explain what it does internally in order to use it with caution, for small content we should use the ByteArray instead (If the PR is merged, I will create a ticket if needed to review all the calls to the related constructors of RsWithBody). But anyway to solve this issue, the only way I found even if I know that it slows down the work of the GC and it is not recommended for this reason, is to remove the file in the finalize method. I guess that you won't like it, but if you have a better idea, please don't hesitate to share it with me, thank you in avance.

essobedo:
@yegor256 Please see my answers and the last version of the PR.

src/main/java/org/takes/rs/Body.java
yegor256: @essobedo wait a second... why do you save the input stream content into a temporary file. this is not really an expected behavior. if I give you an input stream, I don't want your constructor to immediately read it all and save to a file. do I miss something?

yegor256:
@essobedo see above

yegor256:
@krzyk did you/I miss something here? why do we read the entire input stream in a ctor?? this looks very wrong to me

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 please check the description of the PR, I explain everything there, the issue here is the fact that once consumed the InputStream is closed so when you call it again you get an IOException. Moreover sharing an InputStream is incorrect as most of the time they are not thread safe and/or not designed to be shared that is why I have to store the content into a temporary file to be able to share it and read it as many times as you want.

essobedo:
@yegor256 see my answer above and the description of the PR for more details.

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 Actually the thread safety is not even the point here, the point here is more the fact that streams once closed are not supposed to be reused even less to be reused by concurrent threads.

yegor256:
@essobedo I understand your intention, but the design is wrong. your `Body.Stream` class is doing two things at a time: converting `InputStream` to `Body` **and** caching it in a file. That's against the single responsibility principle and is wrong in general. you need two classes, not one.

essobedo:
@yegor256 Here is a new version of the PR that doesn't break the single responsibility principle. please check again

essobedo:
@yegor256 could you please tell me what you think about the new version of the PR? Thank you in advance.

yegor256:
@essobedo it's better now, but check this out: http://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html Constructors must NOT execute any code. Everything should happen in methods of a class. See how it's done in, say, `RqForm.Base`

essobedo:
@yegor256 It should be fine now, could you please check again?

essobedo:
@yegor256 please check again

yegor256:
@essobedo now it looks OK, but I don't like the fact that we're forcing everybody to use this file-caching mechanism. in many/most cases it's not required. if I just want to return my content in input stream, I just do it. take a look at this example: https://github.com/yegor256/rultor/blob/1.61.8/src/main/java/com/rultor/web/TkDaemon.java#L106-L147 You see? Just input stream works perfectly fine. I don't need any files to cache my content. This file-caching is a rather unique feature, which may be needed by someone, but not everybody. See?

essobedo:
@yegor256 unless I miss something but according to what I can see from the code that you point, for me it is not comparable because it creates a new stream at each call (it creates a new instance of SequenceInputStream composed of only new InputStream instances so not shared). In this case, we use a shared InputStream instance so once it has been read and closed it is no more useable as the next calls will fail because the stream is already closed (if we don't close it, we will get -1 at the next calls).
Here I use a file to be able to provide a new InputStream instance at each call like the code that you point, I don't see any better way to do it.

Here is a simple test case that illustrate the issue:
```
   @Test
    public void testStream() throws IOException {
        ByteArrayInputStream is = new ByteArrayInputStream("foo".getBytes());
        SequenceInputStream sis = new SequenceInputStream(Collections.enumeration(Collections.singletonList(is)));
        Assert.assertEquals((int)'f', sis.read());
        Assert.assertEquals((int)'o', sis.read());
        Assert.assertEquals((int)'o', sis.read());
        Assert.assertEquals(-1, sis.read());
        sis.close();
        sis = new SequenceInputStream(Collections.enumeration(Collections.singletonList(is)));
        Assert.assertEquals((int)'f', sis.read());
        Assert.assertEquals((int)'o', sis.read());
        Assert.assertEquals((int)'o', sis.read());
        Assert.assertEquals(-1, sis.read());
    }
```
Here it will fail at the second call because it returns -1 instead of 'f' as a ByteArrayInputStream doesn't care if it is closed or not but with other implementations you will get an IOException. To make it work we would need to reset the underlying stream before starting to read it again but the reset method is not always implemented and moreover it will only allow sequential read not concurrent ones.

essobedo:
@yegor256 Even if it is encapsulated into an immutable object, I don't believe that it is a good practice to share a mutable Object not necessarily Thread Safe such as an InputStream. This PR allows to fix this bad practice.

essobedo:
@yegor256 what do we do from here? any suggestion to go any further?

yegor256:
@essobedo but in most cases we don't need our input stream to be ready for multiple reads. This multiple reads is an exceptional situation. Usually, if I give an input stream to `RsWithBody` I expect it to be read only once. why would anyone be interested in reading it twice? maybe sometimes... but it's an exception, not a default situation

essobedo:
@yegor256 OK so If I understand you correctly for you, doing [this](https://github.com/piotrkot/takesbug/blob/master/src/main/java/com/github/piotrkot/Bug.java#L27-L34) is not a bug in takes but it is rather a misuse of takes, right? As I mention in the ticket itself, in this situation we are supposed to use _Bug.class.getResource_ instead of _Bug.class.getResourceAsStream_, is it what you mean? If so what am I supposed to do within the context of this (false?) bug? I am supposed to simply add a warning in the Javadoc of all constructors of `RsWithBody` with an `InputStream` indicating that it is meant to be read only once so in case it should be used several times we are supposed to use the corresponding constructors with a byte array or a URL. Is it what you expect?

essobedo:
@yegor256 Or maybe you would like me to give access to the constructor `RsWithBody(final Response res, final Body body)` and use `new Body.TempFile(new Body.Stream(body))` only in specific locations such as in `public RsHTML(final InputStream body)`. But for constructors of _RsWithBody_ with `InputStream`, I don't wrap `new Body.Stream(body)` with _Body.TempFile_. Is this what you want?

essobedo:
@yegor256 please check this new version of the PR in which I limited the fix to `RsHTML` and `RsText` in order to use the file-caching approach only on specific use cases as you wanted.
Please note that I had to modify Body.Stream to have the length of the Stream at creation time otherwise I get some failing tests because `this.stream.available()` was called too late so we get IOException instead.

essobedo:
@yegor256 please check again

yegor256:
@essobedo yes, [this](https://github.com/piotrkot/takesbug/blob/master/src/main/java/com/github/piotrkot/Bug.java#L27-L34) is an incorrect use of Takes, not a bug.

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,254 @@
...
+         */
+        private final transient int length;
+
+        /**
+         * Constructs an {@code Stream} with the specified {@link InputStream}.
+         * @param input The content of the body as stream.
+         */
+        Stream(final InputStream input) {
+            this.stream = input;
+            this.length = Body.Stream.length(input);
```
yegor256: @essobedo again, it's wrong. constructors must be code-free! see http://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html

yegor256:
@essobedo one small comment above. the rest is cool

essobedo:
@yegor256 here is a new version of the PR, please check again.

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,252 @@
...
+            this.estimate();
+            return this.length.get();
+        }
+
+        /**
+         * Estimates the length of the {@code InputStream}.
+         * @throws IOException in case the length could not be estimated.
+         */
+        private void estimate() throws IOException {
+            if (this.length.get() == -1) {
```
yegor256: @essobedo what this `if` is for? you do the comparison on the next line

src/main/java/org/takes/rs/Body.java
```diff
@@ -0,0 +1,252 @@
...
+         * @param input The content of the body as stream.
+         */
+        Stream(final InputStream input) {
+            this.stream = input;
+            this.length = new AtomicInteger(-1);
+        }
+
+        @Override
+        public InputStream input() throws IOException {
+            this.estimate();
```
yegor256: @essobedo what is the point of calling it here?

yegor256:
@essobedo a few more questions. sorry for so many comments, but this part of the code is very actively used in many parts of the framework

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 simply to avoid calling `this.stream.available()` each time we call `estimate()`. Since available() is supposed to return a positive value, -1 should be a good choice as initial value to know whether it has already been initialized or not.

src/main/java/org/takes/rs/Body.java
essobedo: @yegor256 because you don't want me to call it in the constructor and in normal use case if we do it in length only we will get an IOException because the stream is already closed (as already explained in previous comment). So to work around this, I have to call estimate in both methods in order to get the value at the first call of one of these methods.

essobedo:
@yegor256 what do you think about my answers?

yegor256:
@rultor try to merge

davvd:
@ypshenychka please, let us know what do you think about this ticket, according to our [QA rules](http://at.teamed.io/qa.html)

davvd:
@rultor deploy pls

ypshenychka:
@davvd Quality is good.

davvd:
> @davvd Quality is good.

@ypshenychka many thanks

davvd:
@krzyk 10 mins was added to the account of @ypshenychka (for QA review), in transaction 81205063. Thanks so much! Your account was topped up for **55 mins** (transaction ID is `81205096`, the task took 411 hours and 16 mins). you're getting extra minutes here (c=40). +55 to your rating, your total score is [+6117](http://www.netbout.com/b/35035?open=rating)

Issue #623, opened 2016-02-26
PsGoogle throws NPE


dmzaytsev:
PsGoogle throws Null pointer  exception 

>Caused by: java.lang.NullPointerException
    at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:199)
    at org.takes.facets.auth.social.PsGoogle.parse(PsGoogle.java:199)
    at org.takes.facets.auth.social.PsGoogle.fetch(PsGoogle.java:151)
    at org.takes.facets.auth.social.PsGoogle.enter(PsGoogle.java:129)
    at org.takes.facets.auth.PsByFlag.enter(PsByFlag.java:107)
    at org.takes.facets.auth.PsChain.enter(PsChain.java:70)
    at com.netbout.rest.PsTwice.enter(PsTwice.java:70)
    at org.takes.facets.auth.TkAuth.act(TkAuth.java:86)
    at org.takes.tk.TkWrap.act(TkWrap.java:61)
    at org.takes.facets.forward.TkForward.act(TkForward.java:68)
    at org.takes.facets.fallback.TkFallback.route(TkFallback.java:85)

see full stack trace https://github.com/yegor256/netbout/issues/1038

PsGoogle.java:199
```
 props.put("name", json.getString("displayName", "unknown"));
```
I guess Google's response doesn't contain `displayName` attribute
the PsGoogle source code should be checked for compliance with Google OAuth specification

Anyway PsGoogle should not throw NPE

dmzaytsev:
@yegor256 could you put urgent tag here?

yegor256:
@davvd it's urgent

davvd:
@dmzaytsev I will find a developer for the task soon...

davvd:
@dmzaytsev there is no milestone yet, so I set it to `1.0`

davvd:
@dmzaytsev many thanks for the bug report, I added 15 mins to your acc, transaction `78661202`

davvd:
@raphaelln This task is yours, please go ahead keeping in mind [this](http://www.xdsd.org/2014/04/17/how-xdsd-is-different.html). If any questions, don't hesitate to ask right here; The budget here is **30 mins**, which is exactly how much time will be paid for, when the task is [completed](http://www.xdsd.org/2014/04/15/definition-of-done.html)

davvd:
> @davvd it's urgent

@yegor256 right, I added "urgent" label

[...]

PR #633, opened 2016-03-01
#623 PsGoogle throws NPE 

raphaelln:
**List of changes**

- **PsGoogle**: Before try to use the json property **displayName**, the program certifies if the displayName is present in the json map, also verify if the **error** property is in the map.
- **PsGoogleTest**: Added 2 junit tests in order to test the missing of displayName property and the presence of error property in the google response.

PS: this correction is just to stop the NPE, we have to open another issue to investigate why the **displayName** is not present in the google response.

### DIFF
--- a/src/main/java/org/takes/facets/auth/social/PsGoogle.java
+++ b/src/main/java/org/takes/facets/auth/social/PsGoogle.java
@@ -195,7 +195,12 @@ public final class PsGoogle implements Pass {
         } else {
             props.put("picture", image.getString("url", "#"));
         }
-        props.put("name", json.getString("displayName", "unknown"));
+        if (json.containsKey("displayName")
+            && json.get("displayName") != null) {
+            props.put("name", json.getString("displayName"));
+        } else {
+            props.put("name", "unknown");
+        }
         return new Identity.Simple(
             String.format("urn:google:%s", json.getString("id")), props
         );
--- a/src/main/java/org/takes/facets/auth/social/PsGoogle.java
+++ b/src/main/java/org/takes/facets/auth/social/PsGoogle.java
@@ -148,11 +148,19 @@ public final class PsGoogle implements Pass {
             .path("me")
             .with("access_token", token)
             .toString();
-        return PsGoogle.parse(
-            new JdkRequest(uri).fetch()
-                .as(JsonResponse.class).json()
-                .readObject()
-        );
+        final JsonObject json = new JdkRequest(uri).fetch()
+            .as(JsonResponse.class).json()
+            .readObject();
+        if (json.containsKey("error")) {
+            throw new HttpException(
+                HttpURLConnection.HTTP_BAD_REQUEST,
+                String.format(
+                    "could not retrieve id from Google, possible cause: %s.",
+                    json.get("message")
+                )
+            );
+        }
+        return PsGoogle.parse(json);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


davvd:
@raphaelln Many thanks, I will find someone to review it before we merge

davvd:
@yegor256 the pull request is rather small, pls review it

src/main/java/org/takes/facets/auth/social/PsGoogle.java
```diff
@@ -195,7 +195,12 @@ private static Identity parse(final JsonObject json) {
         } else {
             props.put("picture", image.getString("url", "#"));
         }
-        props.put("name", json.getString("displayName", "unknown"));
+        if (json.containsKey("displayName")
```
yegor256: @raphaelln I don't think this solves anything. as far as I understand, this method `getString(String,String)` prevents this NULL problem already. if such a property is not there, a default value will be returned (`"unknown"`)

yegor256: @raphaelln try to reproduce the bug in a unit test first. you will see that your solution actually is not a solution

yegor256:
@raphaelln see my comments

raphaelln:
@yegor256 thanks for your review, but my code will prevent the nullpointer in trying to get the **displayName** property.

In the **JsonObjectBuilderImpl** the getString method is:

```
@Override
        public String getString(String name, String defaultValue) {
            try {
                return getString(name);
            } catch (Exception e) {
                return defaultValue;
            }
        }
```

But int the method getString(name):

```
 @Override
        public String getString(String name) {
            return getJsonString(name).getString();  
        }
```

In the *getString(String name) is actually throwing the NPE and the all request is covered by TKFallback (`} catch (final Throwable ex) {`) and the **catch** of the **getString(String name, String defaultValue)** method is not been called because of the TKFallback.

In short, the catch of the *getString(String name, String defaultValue) {* is not been called because the try/catch (final Throwable ex) of the TKFallback.

Also, I believe the error will throw in the **json.getString("id")** either. I think the best resolution of this error is verify the presence of the **error** property in the json object and display the reason to the user.

Following an example of a response with error from google:

```
{
   "error":{
      "errors":[
         {
            "domain":"usageLimits",
            "reason":"accessNotConfigured",
            "message":"Access Not Configured. The API (Google+ API) is not enabled for your project. Please use the Google Developers Console to update your configuration.",
            "extendedHelp":"https://console.developers.google.com"
         }
      ],
      "code":403,
      "message":"Access Not Configured. The API (Google+ API) is not enabled for your project. Please use the Google Developers Console to update your configuration."
   }
}
```

What do you think ?

Thanks

yegor256:
@raphaelln I disagree, and the only way to prove who is right is to create a unit test :)

raphaelln:
@yegor256 I created 2 junit tests:

- Test the presence of the error property in the google response json.
- Test the missing of the displayName in the google response json, the nullpointer after my changes is not occurring anymore.

Please could you review it again ?

Tks.

raphaelln:
@yegor256 please see my comments above.

yegor256:
@rultor try to merge

davvd:
@yegor256 **20 mins** sent to your balance (ID `000-b2e40f29`), many thanks! It took 214 hours and 6 mins.

extra minutes for review comments (c=5)

+20 added to your rating, at the moment it is: [+1822](http://www.netbout.com/b/35008?open=rating)

davvd:
@rultor deploy

PR #649, opened 2016-03-12
Supporting HTTP persistent connections

sebing:
Fixing the issue(https://github.com/yegor256/takes/issues/519)
- Removed  the input stream is closing to single place.
- Make the connection alive until the stream is active.

### DIFF
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -74,11 +74,12 @@ public final class BkBasic implements Back {
     @SuppressWarnings ("PMD.AvoidInstantiatingObjectsInLoops")
     @Override
     public void accept(final Socket socket) throws IOException {
-        final InputStream input = socket.getInputStream();
-        final BufferedOutputStream output = new BufferedOutputStream(
-            socket.getOutputStream()
-        );
-        try {
+        try (
+            final InputStream input = socket.getInputStream();
+            final BufferedOutputStream output = new BufferedOutputStream(
+                socket.getOutputStream()
+            )
+        ) {
             while (true) {
                 this.print(
                     BkBasic.addSocketHeaders(
@@ -91,10 +92,6 @@ public final class BkBasic implements Back {
                     break;
                 }
             }
-        } finally {
-            output.flush();
-            output.close();
-            input.close();
         }
     }
 
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -71,6 +71,7 @@ public final class BkBasic implements Back {
         this.take = tks;
     }
 
+    @SuppressWarnings ("PMD.AvoidInstantiatingObjectsInLoops")
     @Override
     public void accept(final Socket socket) throws IOException {
         final InputStream input = socket.getInputStream();
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -74,15 +74,25 @@ public final class BkBasic implements Back {
     @Override
     public void accept(final Socket socket) throws IOException {
         final InputStream input = socket.getInputStream();
+        final BufferedOutputStream output = new BufferedOutputStream(
+            socket.getOutputStream()
+        );
         try {
-            this.print(
-                BkBasic.addSocketHeaders(
-                    new RqLive(input),
-                    socket
-                ),
-                new BufferedOutputStream(socket.getOutputStream())
-            );
+            while (true) {
+                this.print(
+                    BkBasic.addSocketHeaders(
+                        new RqLive(input),
+                        socket
+                    ),
+                    output
+                );
+                if (input.available() <= 0) {
+                    break;
+                }
+            }
         } finally {
+            output.flush();
+            output.close();
             input.close();
         }
     }
@@ -108,8 +118,6 @@ public final class BkBasic implements Back {
                     HttpURLConnection.HTTP_INTERNAL_ERROR
                 )
             ).print(output);
-        } finally {
-            output.close();
         }
     }
 
--- a/src/test/java/org/takes/http/BkBasicTest.java
+++ b/src/test/java/org/takes/http/BkBasicTest.java
@@ -129,7 +129,7 @@ public final class BkBasicTest {
      *
      * @throws IOException If some problem inside
      * @todo #519:30min In org.takes.rq.RqHeaders.Base#map is skipping the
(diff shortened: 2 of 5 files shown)


davvd:
@sebing Thanks for your pull request, let me find someone who can review it

sebing:
@davvd You did not find someone to review it?

sebing:
@davvd No one reviewing my PR? :(

davvd:
@carlosmiranda this pull request is for you,please review

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -71,18 +71,29 @@ public BkBasic(final Take tks) {
         this.take = tks;
     }
 
+    @SuppressWarnings ("PMD.AvoidInstantiatingObjectsInLoops")
     @Override
     public void accept(final Socket socket) throws IOException {
         final InputStream input = socket.getInputStream();
+        final BufferedOutputStream output = new BufferedOutputStream(
```
carlosmiranda: @sebing let's use try-with-resources feature here since we're now on Java 7. Make sure to include both `input` and `output`.

src/main/java/org/takes/http/BkBasic.java
```diff
@@ -71,18 +71,29 @@ public BkBasic(final Take tks) {
...
+                        socket
+                    ),
+                    output
+                );
+                if (input.available() <= 0) {
+                    break;
+                }
+            }
         } finally {
+            output.flush();
```
carlosmiranda: @sebing No need to call `flush()` here, since `BufferedOutputStream` is an inheritor of `FilterOutputStream` it will be flushed upon close.

src/test/java/org/takes/http/BkBasicTest.java
```diff
@@ -132,7 +128,11 @@ public void exec(final URI home) throws IOException {
      * BkBasic produces headers with addresses without slashes.
      *
      * @throws IOException If some problem inside
+     * @todo #519:30min In org.takes.rq.RqHeaders.Base#map is skipping the
```
carlosmiranda: @sebing How is this related to the current issue? Seems like a new bug to me, let's avoid ignoring unrelated tests if we introduce new bugs, can't we fix it here?

carlosmiranda:
@sebing please see my comments above.

src/test/java/org/takes/http/BkBasicTest.java
sebing: @carlosmiranda I think, I introduce this test failure with my code. I am not sure how i introduce it.
the header is missing `GET /path/c HTTP/1.1`  in the iterator when we run the `addressesInHeadersAddedWithoutSlashes` all other cases it is perfect. 

Any way skipping one item in the iterator here (https://github.com/yegor256/takes/blob/master/src/main/java/org/takes/rq/RqHeaders.java#L142) always and without any comment or instruction why seems a bug  to me.

sebing: @carlosmiranda Should i remove the puzzle and add an issue?

carlosmiranda: @sebing well in this case, your change introduced the bug, better to find out what's wrong and fix it.

At worst, let's create a separate unit test that reproduces this bug specifically, then add a puzzle to fix it and re-enable the tests that were ignored (this test, and your new unit test). The idea is, we want to break as little existing functionality as possible. Based on what I see here, your changes should not have affected this test at all.

sebing: @carlosmiranda I think, I find the problem. 
This is the content we have in the test socket
```
                Joiner.on(BkBasicTest.CRLF).join(
                    "GET / HTTP/1.1",
                    "Host:localhost",
                    "Content-Length: 2",
                    "",
                    "hi"
                ).getBytes()
```

In [RqLive:parse line 99](https://github.com/yegor256/takes/blob/master/src/main/java/org/takes/rq/RqLive.java#L99)
we are breaking the loop. so "hi" remains in the stream. So in the next iteration of BkBasic accept code I have added, try to create the header from just "hi". So the test breaks.

Hope you understood my explanation. Can I add this information too in the puzzle? or should I create a new issue as the `RqLive#parse` is not completely parsing the request. This problem only show up when we want the connection keep alive.

carlosmiranda: @sebing In this case the problem is in `RqLive`, let's create a unit test for it which reproduces the bug, ignore it, and tell the solver of the puzzle to fix the bug and to remove `@Ignore` from that test as well as this one.

sebing: @carlosmiranda Sorry, I have not understood you.
If i remove the `@Ignore` for the test `addressesInHeadersAddedWithoutSlashes` this test will break.
So we have to wait for the new puzzle I create in `RqLive` to finish this task?

carlosmiranda: @sebing No, keep it ignored. Create a new unit test for that specific bug and ignore it as well. In the puzzle instruct the resolver to remove `@Ignore` on _both_ tests after fixing the bug.

Is it clearer now?

sebing: @carlosmiranda Thanks for the clarifications. I will do that.

sebing: @carlosmiranda I think, my diagnostic was again wrong. RqLive separate the header with body with an empty line "". So the code in Rqlive is fine.

sebing:
@carlosmiranda I fixed the test case. Can you re-review the PR?

carlosmiranda:
@sebing looks good, thanks!

carlosmiranda:
@rultor merge please

yegor256:
@rultor try to merge

davvd:
@ypshenychka please, review this issue for QA compliance, as per [par.24](http://at.teamed.io/policy.html#24)

davvd:
@rultor deploy pls

ypshenychka:
@davvd Quality is good.

davvd:
> @davvd Quality is good.

@ypshenychka many thanks

davvd:
@carlosmiranda just added 10 mins to @ypshenychka (for QA), payment ID is `81211238`. **24 mins** was added to your account, many thanks for your contribution (payment 81211268)! The task took 145 hours and 11 mins.. extra minutes for review comments (c=9). added +24 to your rating, now it is equal to [+2575](http://www.netbout.com/b/35023?open=rating)

Issue #655, opened 2016-03-21
RqLive.java:47-49: Refactor the parse method into...


davvd:
Puzzle `519-af7d4687` in `src/main/java/org/takes/rq/RqLive.java:47-49` has to be resolved: Refactor the parse method into a new class. This method is too long and too much if-else statements. Covert this into an IfElseClass.... The puzzle was created by Sebin George on 13-Mar-2016.

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is removed from the source code.

davvd:
@dalifreire it's yours, please go ahead

davvd:
@yegor256 since there is no milestone yet I set it to "1.0"

dalifreire:
@davvd please, assign someone else.

davvd:
> @davvd please, assign someone else.

@dalifreire deducted 30 from your rating :(

davvd:
> @davvd please, assign someone else.

@dalifreire right, I will find someone else, no problem

davvd:
@essobedo please proceed, it's yours

essobedo:
@davvd here is my PR #657 please find a reviewer.

davvd:
> @davvd here is my PR #657 please find a reviewer.

@essobedo OK, sure

essobedo:
@davvd the PR has been merged, please proceed

essobedo:
@davvd please

essobedo:
@davvd I guess that there is a good reason that prevents you to go any further so could you please tell me what it is? Just to allow me to be aware of the entire workflow. Thank you in advance.

essobedo:
@davvd ping

essobedo:
@davvd please close the ticket

davvd:
@yegor256 Thanks, closed

davvd:
@elenavolokhova please, review this ticket for compliance with our [QA rules](http://at.teamed.io/qa.html)

essobedo:
> @yegor256 Thanks, closed
@davvd thanks

elenavolokhova:
@davvd Looks good!

davvd:
> @davvd Looks good!

@elenavolokhova many thanks

davvd:
@essobedo 10 mins was added to the account of @elenavolokhova (for QA review), in transaction 82204435

[...]

PR #657, opened 2016-03-26
#655: Refactor the parse method to reduce the Cyclomatic Complexity

essobedo:
Fix for #655 

The goal of this PR is to reduce the Cyclomatic Complexity of the class RqLive by refactoring the parse method. 

Please note that I did not create a new class as expressed in the puzzle description because the class RqLive is already very simple as it only contains private static methods so creating a new class won't help in this case.

### DIFF
--- a/src/main/java/org/takes/rq/RqLive.java
+++ b/src/main/java/org/takes/rq/RqLive.java
@@ -108,7 +108,7 @@ public final class RqLive extends RqWrap {
      * @param input The input stream to read
      * @param baos Current read header
      * @param position Header line number
-     * @throws IOException If the next byte is not a Line Feed as expected.
+     * @throws IOException If the next byte is not a Line Feed as expected
      */
     private static void checkLineFeed(final InputStream input,
         final ByteArrayOutputStream baos, final Integer position)
--- a/src/main/java/org/takes/rq/RqLive.java
+++ b/src/main/java/org/takes/rq/RqLive.java
@@ -105,6 +105,8 @@ public final class RqLive extends RqWrap {
 
     /**
      * Checks whether or not the next byte to read is a Line Feed.
+     * <p><i>Please note that this method assumes that the previous byte read
+     * was a Carriage Return.</i>
      * @param input The input stream to read
      * @param baos Current read header
      * @param position Header line number
--- a/src/main/java/org/takes/rq/RqLive.java
+++ b/src/main/java/org/takes/rq/RqLive.java
@@ -43,13 +43,8 @@ import org.takes.misc.Opt;
  * @author Yegor Bugayenko (yegor@teamed.io)
  * @version $Id$
  * @since 0.1
- * @checkstyle CyclomaticComplexityCheck (500 lines)
- * @todo #519:30min Refactor the parse method into a new class.
- *  This method is too long and too much if-else statements. Covert this
- *  into an IfElseClass.
  */
 @EqualsAndHashCode(callSuper = true)
-@SuppressWarnings("PMD.CyclomaticComplexity")
 public final class RqLive extends RqWrap {
 
     /**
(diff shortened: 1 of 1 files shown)


davvd:
@essobedo Many thanks for the PR, let me find a reviewer for it

davvd:
@mkordas review this one, please

mkordas:
@essobedo I'll do a review today

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -132,6 +104,28 @@ public InputStream body() {
...
+     * @throws IOException If the next byte is not a Line Feed as expected.
+     */
+    private static void checkLineFeed(final InputStream input,
+        final ByteArrayOutputStream baos, final Integer position)
+        throws IOException {
+        if (input.read() != '\n') {
+            throw new HttpException(
+                HttpURLConnection.HTTP_BAD_REQUEST,
+                String.format(
+                    "there is no LF after CR in header, line #%d: \"%s\"",
```
mkordas: @essobedo this method should be more generic, we can assume it will always be invoked after CR

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -181,16 +174,20 @@ private static Integer legalCharacter(final Opt<Integer> data,
...
      * @return Next or current data
      * @throws IOException if input.read() fails
      */
     private static Opt<Integer> data(final InputStream input,
-        final Opt<Integer> data) throws IOException {
+        final Opt<Integer> data, final boolean available) throws IOException {
         final Opt<Integer> ret;
         if (data.has()) {
             ret = data;
+        } else if (available && input.available() <= 0) {
```
mkordas: @essobedo do we have tests for this new `if` here?

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -68,53 +63,30 @@ public RqLive(final InputStream input) throws IOException {
...
                 if (baos.size() == 0) {
                     break;
                 }
-                data = new Opt.Single<Integer>(input.read());
+                data = new Opt.Single<>(input.read());
                 final Opt<String> header = RqLive.newHeader(data, baos);
                 if (header.has()) {
                     head.add(header.get());
                 }
+                data = RqLive.data(input, data, false);
```
mkordas: @essobedo what do we refactor by putting this line here?

src/main/java/org/takes/rq/RqLive.java
```diff
@@ -132,6 +104,28 @@ public InputStream body() {
     }
 
     /**
+     * Checks whether or not the next byte to read is a Line Feed.
+     * @param input The input stream to read
+     * @param baos Current read header
+     * @param position Header line number
+     * @throws IOException If the next byte is not a Line Feed as expected.
```
mkordas: @essobedo we don't need that dot at the end here

mkordas:
@essobedo please see my comments

src/main/java/org/takes/rq/RqLive.java
essobedo: @mkordas could you please explain why it should be more generic? Here is a private static method called by another private static method so it is internal to the class RqLive and used for a very specific use case it is not meant to be reused elsewhere even internally. If for another task it is needed to make it generic, it can be done for this particular task, here our goal is to do the task in the most efficient manner, don't you agree?

src/main/java/org/takes/rq/RqLive.java
essobedo: @mkordas it is not easy to test it directly, but if you remove it you will currently get 6 failing tests so yes it is unit tested but indirectly. The other aspect to note is the fact that even if I refactored the code, I strictly respected the previous method flow because if you don't you will get failing tests.

src/main/java/org/takes/rq/RqLive.java
essobedo: @mkordas same answer as above, if you check well I strictly respected the previous method flow because if you don't you will get failing tests.

src/main/java/org/takes/rq/RqLive.java
essobedo: @mkordas ok

essobedo:
@mkordas thx for your remarks, please see my answers above. I removed the useless dot in the last version of the PR

src/main/java/org/takes/rq/RqLive.java
mkordas: @essobedo this is kind of http://www.yegor256.com/2015/12/08/temporal-coupling-between-method-calls.html. We need to invoke this method only after CR, otherwise it will print message that is not true. I just propose in this case, to put info about CR in Javadoc, this should be enough

src/main/java/org/takes/rq/RqLive.java
mkordas: @essobedo that's enough, thanks

src/main/java/org/takes/rq/RqLive.java
mkordas: @essobedo thanks

mkordas:
@essobedo just one open issue is left

src/main/java/org/takes/rq/RqLive.java
essobedo: @mkordas makes perfect sense thx for the remark.

essobedo:
@mkordas PR updated, please check again

mkordas:
@essobedo looks good!

mkordas:
@rultor merge

mkordas:
@yegor256 we need your review here

davvd:
@ypshenychka please, review this ticket for compliance with our [QA rules](http://at.teamed.io/qa.html)

davvd:
@rultor deploy pls

ypshenychka:
@mkordas 
According to our [QA Rules](http://at.teamed.io/qa.html):

Only two issues (one of which was cosmetic) were found during code review. 
Please confirm that you'll try to find at least three major problems along with cosmetic concerns while future reviews.

mkordas:
@ypshenychka I confirm

ypshenychka:
@mkordas Thank you.

ypshenychka:
@davvd Quality is acceptable.

davvd:
> @davvd Quality is acceptable.

@ypshenychka thanks for the QA review, we'll work better next time

davvd:
@mkordas I added 10 mins to @ypshenychka (for QA review) in transaction 82174184. **29 mins** was added to your account, many thanks for your contribution (payment 56ffa431d553871dea0002ff)! The task took 78 hours and 27 mins.. you're getting extra minutes here (c=14). +29 to your rating, your total score is [+9116](http://www.netbout.com/b/36080?open=rating)

Issue #660, opened 2016-04-12
pom.xml:338-340: Continue fixing FindBugs violations started...


davvd:
Puzzle `624-78dfe52a` in `pom.xml:338-340` has to be resolved: Continue fixing FindBugs violations started in #624. #624 turned on the FindBugs, temporarily excluding some classes. Remove all excludes of FindBugs below and fix the issues.... The puzzle was created by Dalí Freire on 29-Mar-2016.

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is removed from the source code.

<hr id='w'>

davvd:
@essobedo please proceed, it's yours

davvd:
@yegor256 I set milestone here to `1.0`, let me know if it is wrong

essobedo:
@davvd here is my PR #662 please find a reviewer

davvd:
> @davvd here is my PR #662 please find a reviewer

@essobedo OK, sure

essobedo:
@davvd waiting for https://github.com/teamed/qulice/issues/762. Indeed I need some help from the Qulice Team to fix it properly

essobedo:
@davvd could you please confirm that you approve that this issue depends on https://github.com/teamed/qulice/issues/762 ?

essobedo:
@davvd waiting for https://github.com/teamed/qulice/issues/762

essobedo:
@davvd @yegor256 this ticket depends on https://github.com/teamed/qulice/issues/762 so we should have "Impediment" in the details of this ticket in my agenda, shouldn't we?

essobedo:
@davvd need more time to get the required information from https://github.com/teamed/qulice/issues/762

davvd:
> @davvd need more time to get the required information from https://github.com/teamed/qulice/issues/762

@essobedo of course, thanks for letting me know, take your time

essobedo:
@davvd FYI https://github.com/teamed/qulice/issues/762 is now closed, the PR has been approved by the reviewer, so now we only wait for the architect to review it

essobedo:
@davvd need more time to be validated by the architect

davvd:
> @davvd need more time to be validated by the architect

@essobedo sure, take your time, I know about this pause

essobedo:
@davvd the PR has been merged, you can proceed, thx

davvd:
[...]

PR #662, opened 2016-04-13
#660: Fixed FindBugs violations

essobedo:
Fix for #660 

The goal of this PR is to fix the last existing FindBugs violations. 

**NB:** To be able to "fix" 2 issues related to deserialization (aka _SE_TRANSIENT_FIELD_NOT_RESTORED: Transient field that isn't set by deserialization_), I had to suppress a warning for each of them: 

- For the class _HttpException_, I fixed the issue by removing the modifier _transient_ but I get a warning from PMD (_BeanMembersShouldSerialize_ ) complaining about the fact that I don't have any accessor which is not true here since I have the method _code()_ instead of the expected method _getStatus()_ so I suppressed the warning "PMD.BeanMembersShouldSerialize".
- For the class _RsForward_, I didn't find any obvious way to fix it since the variable _origin_ is not necessary _Serializable_ so it needs to remain _transient_ so I added the corresponding annotation SuppressFBWarnings and to do so I added a new _provided_ dependency to the pom file.


davvd:
@essobedo Let me find a reviewer for this pull request, thanks for submitting it

davvd:
@dmzaytsev review this one,please

dmzaytsev:
@essobedo I would suggest you submit a ticket in Qulice project and ask how the FindBugs violations can be fixed instead of the supressions. I mean RsForward and HttpException.
If Qulice guys will confirm the ticket as a bug, you could create a puzzle to remove the suppressions in `pom.xml` later

essobedo:
@dmzaytsev thank you for the suggestion, here is my ticket https://github.com/teamed/qulice/issues/762.   What about the rest of the PR do you have any remarks?

dmzaytsev:
@essobedo the rest of PR looks good, thanks

dmzaytsev:
@essobedo please create a new puzzle to remove the suppression in HttpException once the next release of Qulice will be available
and we will merge it as is

essobedo:
@dmzaytsev done please check again

dmzaytsev:
@essobedo thanks

dmzaytsev:
@rultor merge

essobedo:
@yegor256 please check

essobedo:
@yegor256 ping

essobedo:
@yegor256 Please confirm

essobedo:
@yegor256 your validation is needed here

src/main/java/org/takes/facets/forward/TkForward.java
```diff
@@ -107,21 +106,19 @@ public InputStream body() throws IOException {
...
-                    InputStream body;
-                    try {
-                        head = this.origin.head();
-                        body = this.origin.body();
-                    } catch (final RsForward ex) {
-                        head = ex.head();
-                        body = ex.body();
-                    }
-                    this.saved.add(new RsSimple(head, body));
+            if (this.saved.get() == null) {
```
yegor256: @essobedo this code is NOT thread-safe

src/main/java/org/takes/http/MainRemote.java
```diff
@@ -83,7 +83,9 @@ public MainRemote(final Class<?> type, final String... passed) {
      */
     public void exec(final MainRemote.Script script) throws Exception {
         final File file = File.createTempFile("takes-", ".txt");
-        file.delete();
+        if (!file.delete()) {
+            file.deleteOnExit();
```
yegor256: @essobedo this looks strange... why not throwing an exception?

src/main/java/org/takes/http/MainRemote.java
```diff
@@ -104,7 +106,9 @@ public void exec(final MainRemote.Script script) throws Exception {
                 )
             );
         } finally {
-            file.delete();
+            if (!file.delete()) {
```
yegor256: @essobedo this looks strange... why not throwing an exception?

src/main/java/org/takes/rq/RqForm.java
```diff
@@ -156,12 +157,13 @@ private static String decode(final CharSequence txt) {
          * @throws IOException If something fails reading or parsing body
          */
         private Map<String, List<String>> map() throws IOException {
-            synchronized (this.saved) {
-                if (this.saved.isEmpty()) {
-                    this.saved.add(this.freshMap());
-                }
-                return this.saved.get(0);
+            if (this.saved.get() == null) {
```
yegor256: @essobedo this code is not thread-safe

src/main/java/org/takes/rs/RsGzip.java
```diff
@@ -82,21 +81,20 @@ public InputStream body() throws IOException {
...
-                        new RsWithBody(
-                            this.origin,
-                            RsGzip.gzip(this.origin.body())
-                        ),
-                        "Content-Encoding",
-                        "gzip"
-                    )
-                );
-            }
+        if (this.zipped.get() == null) {
```
yegor256: @essobedo same here, it's not thread-safe

src/main/java/org/takes/rs/RsPrettyJSON.java
```diff
@@ -87,17 +86,16 @@ public InputStream body() throws IOException {
...
-        synchronized (this.transformed) {
-            if (this.transformed.isEmpty()) {
-                this.transformed.add(
-                    new RsWithBody(
-                        this.origin,
-                        RsPrettyJSON.transform(this.origin.body())
-                    )
-                );
-            }
+        if (this.transformed.get() == null) {
```
yegor256: @essobedo same here

yegor256:
@essobedo see my comments above

yegor256:
@dmzaytsev didn't you see that the new code is not thread safe?

src/main/java/org/takes/http/MainRemote.java
essobedo: @yegor256 because I don't want to change the current method flow, for now it assumes that even if we cannot delete the file, it is not important for the rest of the method. But If you prefer an Exception, I can throw it

src/main/java/org/takes/http/MainRemote.java
essobedo: @yegor256 same answer as above

essobedo:
@yegor256 thx for the remarks, I fixed the thread safety issues, please check again

essobedo:
@yegor256 please check again

essobedo:
@yegor256 please check again

essobedo:
@yegor256 ping

src/main/java/org/takes/facets/forward/TkForward.java
```diff
@@ -107,21 +111,23 @@ public InputStream body() throws IOException {
...
-                    InputStream body;
-                    try {
-                        head = this.origin.head();
-                        body = this.origin.body();
-                    } catch (final RsForward ex) {
-                        head = ex.head();
-                        body = ex.body();
+            if (this.saved.get() == null) {
+                synchronized (this.mutex) {
+                    if (this.saved.get() == null) {
```
yegor256: @essobedo doesn't it smell for you? looks rather strange. what's wrong about `synchronized` block we had before? why do we need to refactor it?

src/main/java/org/takes/http/MainRemote.java
yegor256: @essobedo definitely an exception, nothing else, see http://www.yegor256.com/2015/08/25/fail-fast.html

src/main/java/org/takes/http/MainRemote.java
yegor256: @essobedo same here, see http://www.yegor256.com/2015/08/25/fail-fast.html

src/main/java/org/takes/rq/RqForm.java
```diff
@@ -156,12 +162,16 @@ private static String decode(final CharSequence txt) {
          * @throws IOException If something fails reading or parsing body
          */
         private Map<String, List<String>> map() throws IOException {
-            synchronized (this.saved) {
-                if (this.saved.isEmpty()) {
-                    this.saved.add(this.freshMap());
+            if (this.saved.get() == null) {
+                synchronized (this.mutex) {
+                    if (this.saved.get() == null) {
```
yegor256: @essobedo same here, it's definitely wrong

yegor256:
@essobedo see my comments above

src/main/java/org/takes/facets/forward/TkForward.java
essobedo: @yegor256 because `CopyOnWriteArrayList` is already thread safe so synchronizing access to a thread safe object is considered as a violation by FindBugs as it sounds like a misuse but if you prefer that I put it back with and Ignore I can, even if I personally find the previous approach much worse than the one I proposed, indeed using a `CopyOnWriteArrayList` to add only one element and return the first element sounds really like an ugly hack to me, it is only my point of view. The only advantage of the old approach compared to mine is the fact that you don't pay the price of an access to a `volatile` variable so it will be slightly faster but that's all. So what should I do? Revert the changes and ignore it?

src/main/java/org/takes/rq/RqForm.java
essobedo: @yegor256 You can dislike my approach but I don't understand why it is wrong, it is thread safe and does the exact same thing with more appropriate objects IMHO. It is the well known double checking approach based on a `volatile` variable, I know that it is a controversy approach but it is not wrong

src/main/java/org/takes/http/MainRemote.java
essobedo: @yegor256 ok let's throw an exception then

src/main/java/org/takes/http/MainRemote.java
essobedo: @yegor256 ok let's throw an exception then

essobedo:
@yegor256 I reverted my changes to fix [JLM_JSR166_UTILCONCURRENT_MONITORENTER](http://findbugs.sourceforge.net/bugDescriptions.html#JLM_JSR166_UTILCONCURRENT_MONITORENTER) and now I throw an IOException if the file could not be deleted. As we cannot throw an IOException in a finally block I decided to slip it in 2: In case of an exception I add it to suppressed exceptions otherwise I throw IOException outside the finally block

essobedo:
@yegor256 ping

essobedo:
@yegor256 ping

essobedo:
@yegor256 please we need to do something with this PR, it is open for more than 3 weeks now

src/main/java/org/takes/http/MainRemote.java
```diff
@@ -49,6 +49,12 @@
 public final class MainRemote {
 
     /**
+     * The format of the error message indicating that the file could not be
+     * deleted.
+     * @checkstyle LineLength (2 lines)
+     */
+    private static final String DEL_FILE_ERR_MSG = "The file '%s' could not be deleted";
```
yegor256: @essobedo brrr... check this out: http://www.yegor256.com/2016/05/03/test-methods-must-share-nothing.html not exactly the case, but the problem is the same - why the text is exactly the same? these two pieces of code are connected somehow?

yegor256:
@essobedo one more comment above

src/main/java/org/takes/http/MainRemote.java
essobedo: @yegor256 because it is the exact same error, exact same context, exact same method even exact same file so for me it has nothing to do with your blog post that I know already BTW, but still OK will try to give different messages anyway.

essobedo:
@yegor256 fixed, please check again

essobedo:
@yegor256 ping

essobedo:
@yegor256 let's finish it please

yegor256:
@rultor try to merge

essobedo:
@yegor256 thank you

davvd:
@elenavolokhova please, review this issue for QA compliance, as per [par.24](http://at.teamed.io/policy.html#24)

davvd:
@rultor please deploy

elenavolokhova:
@dmzaytsev 
 > The code reviewer found at least three problems in the code.
 
 Only one issue was found during review.
Please confirm that you will try to find at least three problems next time.

dmzaytsev:
@elenavolokhova I confirm

elenavolokhova:

davvd:

@elenavolokhova thanks for the QA review, we'll work better next time

davvd:
@dmzaytsev just added 10 mins to @elenavolokhova (for QA), payment ID is `86614791`. thanks, paid, **21 mins** to your account, payment ID is `86614802`, task took 677 hours and 9 mins to complete. extra minutes for review comments (c=6). +21 added to your rating, at the moment it is: [+1027](http://www.netbout.com/b/35166?open=rating)

Issue #812, opened 2018-02-26
RsPrintTest.java:75-78: Fix...


0pdd:
The puzzle `778-192143c0` from #778 has to be resolved:

https://github.com/yegor256/takes/blob/55b67f2975dd00a9ce21fd0862ed650d5556e64b/src/test/java/org/takes/rs/RsPrintTest.java#L75-L78

The puzzle was created by rultor on 26-Feb-18. 

Estimate: 30 minutes, role: DEV.

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@yegor256[/z](https://www.0crat.com/u/yegor256) please, pay attention to this issue

0crat:
Job [#812](https://github.com/yegor256/takes/issues/812) is now in scope, role is `DEV`

0crat:
The job [#812](https://github.com/yegor256/takes/issues/812) assigned to @krzyk[/z](https://www.0crat.com/u/krzyk), here is [why](https://www.0crat.com/footprint/C3FFK3YAY/2501004776); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

0crat:
The job [#812](https://github.com/yegor256/takes/issues/812) assigned to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso), here is [why](https://www.0crat.com/footprint/C3FFK3YAY/2562346536); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

[...]

PR #870, opened 2018-11-20
Fix flushHeadEvenWhenExceptionHappens in RsPrintTest

victornoel:
This is for #812.

Because `OutputStreamWriter` is buffered, the `write` method which was throwing the test exception was called only when `flush()` was called and this prevented the test to validate its premises.

I remove the need for `OutputStreamWriter` in the test to solve this situation.

This is coherent with the fact that printing the header has to be character oriented as with a `Writer`.


0crat:
Job [#870](https://github.com/yegor256/takes/issues/870) is now in scope, role is `REV`

src/main/java/org/takes/rs/RsPrint.java
```diff
@@ -115,41 +115,52 @@ public void print(final OutputStream output) throws IOException {
...
+                }
+                writer.append(line);
+                writer.append(eol);
+                ++pos;
             }
-            writer.append(line);
             writer.append(eol);
-            ++pos;
+        } finally {
+            writer.flush();
```
g4s8: @victornoel the ticket #812 says:
> Implementation could not rely 
>  on simple try-finally block as OutputStreamWriter do not call 
>  to the flush() of OutputStream. It has additional checks. 

so this fix goes against requested changes.

src/test/java/org/takes/rs/RsPrintTest.java
```diff
@@ -72,30 +73,77 @@ public void flushBodyEvenWhenExceptionHappens() throws IOException {
...
-     *  and remove Ignore annotation. Implementation could not rely
-     *  on simple try-finally block as OutputStreamWriter do not call
-     *  to the flush() of OutputStream. It has additional checks.
      */
     @Test
-    @Ignore
     public void flushHeadEvenWhenExceptionHappens() throws IOException {
         final IOException exception = new IOException("Error");
-        final FailOutputStream output = new FailOutputStream(exception);
+        final FailWriter writer = new FailWriter(exception);
```
g4s8: @victornoel why can't we keep this test as is? Changing the test may miss some potential failures, which were covered with previous test.

src/main/java/org/takes/rs/RsPrint.java
```diff
@@ -115,41 +115,52 @@ public void print(final OutputStream output) throws IOException {
...
 
     /**
-     * Print it into output stream.
+     * Print it into output stream in UTF8.
      * @param output Output to print into
      * @throws IOException If fails
      * @since 0.10
      */
     public void printHead(final OutputStream output) throws IOException {
+        this.printHead(new Utf8OutputStreamWriter(output));
```
g4s8: @victornoel ~~I think it's not correct to enforce encoding to `UTF-8`, user may configure encoding via `-Dfile.encoding` JVM argument, so it's better to use `Charset.defaultCharset()` for encoding.~~

Update: it's not related to this PR, because same approach was used previously, so I've submitted a ticket: #871

src/main/java/org/takes/rs/RsPrint.java
victornoel: @g4s8 no it does not, as I explained in the PR description, the problem was that the test didn't work because `write` (which throws an exception for the sake of the test) was called during the `finally` phase of the try-finally block instead of inside the `try`.
I solved that situation (which was caused only by the test, not when using `RsPrint` for real) so now it is not the case anymore and it is working as expected.

paulodamaso:
@rultor merge

0crat:
@ypshenychka[/z](https://www.0crat.com/u/ypshenychka) please review this job completed by @g4s8[/z](https://www.0crat.com/u/g4s8), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

0crat:
The job [#870](https://github.com/yegor256/takes/issues/870) is now out of scope

0crat:
Payment to `ARC` for a closed pull request, as in [§28](http://www.zerocracy.com/policy.html#28): +10 point(s) just awarded to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso)

ypshenychka:
@g4s8 According to our [QA Rules](http://www.zerocracy.com/policy.html#42):

No issues were found during code review. 
Please confirm that you'll try to find at least three major problems while future reviews.

g4s8:
@ypshenychka I confirm

Issue #876, opened 2018-11-28
PsFacebook seems not to support UTF8 content


victornoel:
Since PsFacebookTest relies on a random id to do its test, sometimes it retrieves UTF8 content which can't be parsed by Takes and/or restfb:
```
canLogin(org.takes.facets.auth.social.PsFacebookTest)  Time elapsed: 0.056 sec  <<< ERROR!
java.lang.IllegalArgumentException: com.restfb.exception.FacebookJsonMappingException: Unable to map JSON to Java. Offending JSON is '{"id":"JD6U6vubd1","name":"ⲉ鯨ɯ摓㻦恛҉祐፲"}'.
	at org.takes.facets.auth.social.PsFacebook.fetch(PsFacebook.java:199)
	at org.takes.facets.auth.social.PsFacebook.enter(PsFacebook.java:157)
	at org.takes.facets.auth.social.PsFacebookTest.canLogin(PsFacebookTest.java:83)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.apache.maven.surefire.junitcore.pc.Scheduler$1.run(Scheduler.java:393)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #885, opened 2018-12-07
#876 corrected test for PsFacebook

krzyk:
#876 
* corrected test for PsFacebook to generate correct unicode characters (and not control sequences - `random()` was generating those and that's why the test was sometimes failing)

### DIFF
--- a/src/test/java/org/takes/facets/auth/social/PsFacebookTest.java
+++ b/src/test/java/org/takes/facets/auth/social/PsFacebookTest.java
@@ -26,9 +26,10 @@ package org.takes.facets.auth.social;
 import com.jcabi.http.request.FakeRequest;
 import com.restfb.DefaultWebRequestor;
 import java.net.HttpURLConnection;
+import java.nio.charset.StandardCharsets;
 import java.util.Collections;
-import java.util.Map;
 import org.apache.commons.lang.RandomStringUtils;
+import org.apache.commons.text.RandomStringGenerator;
 import org.hamcrest.CoreMatchers;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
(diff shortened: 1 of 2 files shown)


0crat:
Job [#885](https://github.com/yegor256/takes/issues/885) is now in scope, role is `REV`

krzyk:
@paulodamaso no reviewer was assigned here

src/test/java/org/takes/facets/auth/social/PsFacebookTest.java
```diff
@@ -52,15 +53,20 @@
     @Test
     public final void canLogin() throws Exception {
         final String identifier = RandomStringUtils.randomAlphanumeric(10);
+        final RandomStringGenerator generator =
```
paulodamaso: @krzyk I think that we could replace `RandomStringGenerator` and `RandomStringUtils` usage with `RandomText` from cactoos, WDYT?

krzyk: @paulodamaso But `RandomText` configuration is cumbersome, I would need to list all UTF characters there and do filtering, all this is done automatically byt RandomStringGenerator and RandomStringUtils.

paulodamaso: @krzyk `RandomText` comes with an pre-defined set of characters, so you'd just need to filter the unwanted ones, I think that it'll be enough for our test cases, WDYT?

krzyk: @paulodamaso OK, and are you OK with IntStream for generating the list of characters?

paulodamaso: @krzyk Yes, there is something that should I be worried about?

krzyk: @paulodamaso No, I was worried that I would need to embed zillions of classes from cactoos (because there is no direct way to generate a List or Iterator with provided seed and function) to get the same thing I can do in 4-5 lines with streams :)

krzyk: @paulodamaso So after reviewing it more, it is impossible to do. RandomText should accept a list of codepoints, but it accepts only Characters. One codepoint can result in mutliple chars (and the order is important here). And here we want to test unicode chars that can be outside of the simple 64k space that can be represented by a single char.

paulodamaso: @krzyk Okay, so that's nothing else to do here, thanks!

paulodamaso:
@krzyk Please just see why your build is failing on travis

krzyk:
@paulodamaso done

paulodamaso:
@rultor merge

0crat:
Job `gh:yegor256/takes#885` is not assigned, can't get performer

0crat:
The job [#885](https://github.com/yegor256/takes/issues/885) is now out of scope

Issue #847, opened 2018-09-05
Add "How to contribute" session in README.md to Gradle users


paulodamaso:
Continuing #790 work, add instructions for building and testing Takes with Gradle instead of Maven including the execution of all integration tests and qulice profile for static analysis.

0crat:
@yegor256[/z](https://www.0crat.com/u/yegor256) please, pay attention to this issue

paulodamaso:
0crat in

paulodamaso:
@0crat status

0crat:
> @0crat status ([here](https://github.com/yegor256/takes/issues/847#issuecomment-430977753))

@paulodamaso This is what I know about this job in [C3FFK3YAY](https://www.0crat.com/p/C3FFK3YAY), as in [§32](http://www.zerocracy.com/policy.html#32):

  * The job [#847](https://github.com/yegor256/takes/issues/847) is [in scope](http://www.zerocracy.com/policy.html#14) for 21s
  * The role is `DEV`
  * The job is not assigned to anyone
  * The [budget](http://www.zerocracy.com/policy.html#4) is not set yet
  * Job [footprint](https://www.0crat.com/footprint/C3FFK3YAY?q=%7Bjob%3A%27gh%3Ayegor256%2Ftakes%23847%27%2Ctype%3A%7B%24not%3A%2F%5ENotify%2F%7D%7D) (restricted area)

0crat:
Bug was reported, see [§29](http://www.zerocracy.com/policy.html#29): +15 point(s) just awarded to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso)

0crat:
The job [#847](https://github.com/yegor256/takes/issues/847) assigned to @victornoel[/z](https://www.0crat.com/u/victornoel), here is [why](https://www.0crat.com/footprint/C3FFK3YAY/1bbf81b0-29d7-4463-bc2a-1e962d904402); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

victornoel:
@0crat refuse

0crat:
[...]

PR #919, opened 2018-12-27
#847 Add "How to contribute" session in README.md to Gradle users

fabriciofx:
@paulodamaso I've started to add Gradle support to Takes. However, Gradle there are some problems using it, like errors in tests (20) and need add support to use Qulice too. These new tasks will need add them like PDD puzzles, but I don't know how do it in documentation or gradle configuration file. What I should do?


0crat:
Job [#919](https://github.com/yegor256/takes/issues/919) is now in scope, role is `REV`

paulodamaso:
@fabriciofx Thanks for the PR, but thinking better about this issue looks like we don't need another build / dependency management tool besides maven. Adding gradle support would make us to add all or CI tools to work with gradle too, duplicating them, and I don't think it is worth it.

0crat:
Job `gh:yegor256/takes#919` is not assigned, can't get performer

0crat:
The job [#919](https://github.com/yegor256/takes/issues/919) is now out of scope

Issue #1055, opened 2020-06-29
RqPrint.java:41-43: This class should implement Text in...


0pdd:
The puzzle `984-b6700210` from #984 has to be resolved: 

https://github.com/yegor256/takes/blob/ef29898cc602dcffc7a52498b59cce7dbdea3917/src/main/java/org/takes/rq/RqPrint.java#L41-L43

The puzzle was created by andreoss on 23-Jun-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

0pdd:
The puzzle `984-b6700210` has disappeared from the source code, that's why I closed this issue.

0crat:
Job `gh:yegor256/takes#1055` is not assigned, can't get performer

PR #1056, opened 2020-06-30
For #1055: Implement Text by RqPrint

andreoss:
See #1055

### DIFF
--- a/src/main/java/org/takes/rq/RqPrint.java
+++ b/src/main/java/org/takes/rq/RqPrint.java
@@ -41,6 +41,7 @@ import org.takes.Request;
  *
  * <p>The class is immutable and thread-safe.
  * @since 0.1
+ * @checkstyle ClassDataAbstractionCouplingCheck (200 lines)
  */
 @EqualsAndHashCode(callSuper = true)
 public final class RqPrint extends RqWrap implements Text {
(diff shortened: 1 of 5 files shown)


andreoss:
@paulodamaso Have a look, please

src/main/java/org/takes/rq/RqPrint.java
```diff
@@ -29,30 +29,50 @@
...
- *
  * @since 0.1
  */
 @EqualsAndHashCode(callSuper = true)
-public final class RqPrint extends RqWrap {
+public final class RqPrint extends RqWrap implements Text {
+
+    /**
+     * The textual representation.
+     * @todo #1050:30m When (and if) cactoos/#1403 is done, change type of this
```
paulodamaso, line 50: @andreoss Seems that we won't have cactoos#1403 done, but I think we might use `Text` here anyway, WDYT?

andreoss: @paulodamaso Done with anonymous class. The problem here is that `TextOf` does not have a public ctor for `Scalar`, and there is no decorator which makes `Text` compute its value once.

paulodamaso:
@rultor merge

0crat:
Job `gh:yegor256/takes#1056` is not assigned, can't get performer

Issue #1036, opened 2020-03-09
Fix Rultor build


paulodamaso:
Fix rultor build / merge / release

paulodamaso:
@rultor release tag is `1.20`

paulodamaso:
@rultor release tag is `1.20`

PR #1066, opened 2020-09-03
For #1036: removing pdd and xcop installation from rultor file, alrea…

paulodamaso:
For #1036: removing pdd and xcop installation from rultor file, already installed in used rultor image


0crat:
Job `gh:yegor256/takes#1066` is not assigned, can't get performer

Issue #1095, opened 2021-03-21
RqMtBase.java:128-133: Remove code from this ctor,...


0pdd:
The puzzle `950-0b2d9fd8` from #950 has to be resolved: 

https://github.com/yegor256/takes/blob/925e1a2cdef20afecb4b55a650a6b749dc390b1c/src/main/java/org/takes/rq/multipart/RqMtBase.java#L128-L133

The puzzle was created by @andreoss on 22-Oct-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `950-0b2d9fd8` has disappeared from the source code, that's why I closed this issue.

PR #1098, opened 2021-04-01
(#1095) Remove code from ctor

baudoliver7:
For #1095

- Change code in ctor to `Sticky` initialization
- Remove lazy loading evaluation in `RqMtFake`


andreoss:
@baudoliver7 The tests are failing

pom.xml
```diff
@@ -532,4 +528,4 @@ SOFTWARE.
       </build>
     </profile>
   </profiles>
-</project>
+</project>
```
andreoss, line 531: @baudoliver7 I suggest reverting this part, if xcop doesn't complain than xml formating is fine

src/main/java/org/takes/rq/multipart/RqMtBase.java
```diff
@@ -125,27 +127,24 @@
...
      */
     public RqMtBase(final Request req) throws IOException {
         this.origin = req;
-        this.stream = new RqLengthAware(req).body();
-        this.buffer = ByteBuffer.allocate(
-            // @checkstyle MagicNumberCheck (1 line)
-            Math.min(8192, this.stream.available())
-        );
-        this.map = this.requests(req);
+        this.sstream = new Sticky<>(() -> new RqLengthAware(req).body());
```
andreoss, line 134: @baudoliver7 Is it possible to use `RqOnce` on the ctor parameter instead of three Stickies? 
It seems also that some fields could be moved inside of `request` method, such as `sbuffer`

andreoss (changes requested):

yegor256:
@baudoliver7 what's up with this one?

yegor256:
Please fix the merge conflicts so this pull request can be merged.

baudoliver7:
@yegor256 It is good now :)

Issue #1118, opened 2022-03-16
Bug while trying to authenticate with PsGithub


baudoliver7:
We get this message :
```console
Must specify access token via Authorization header. https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param
```

Url for documentation :

- https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/
- https://docs.github.com/v3/#oauth2-token-sent-in-a-header

baudoliver7:
@yegor256 please, assign

PR #1119, opened 2022-03-18
fix(github): fix deprecating auth through query param

baudoliver7:
Now, Github only allows sending token in the header. We fixed it.
See: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

Fix: #1118


baudoliver7:
@yegor256 please, review and merge
A few services based on `PsGithub` need this fix (`Zerocracy`, `Rultor`, etc.)

yegor256:
@baudoliver7 thanks!

PR #1122, opened 2022-04-27
Use `cactoos.InputOf`

Masynchin:
Sorry, but I haven't run any tests. I'm not Java programmer, but I think this changes makes this code more Java-way.

### DIFF
--- a/src/main/java/org/takes/tk/TkFiles.java
+++ b/src/main/java/org/takes/tk/TkFiles.java
@@ -24,9 +24,7 @@
 package org.takes.tk;
 
 import java.io.File;
-import java.io.IOException;
 import java.net.HttpURLConnection;
-import java.nio.file.Files;
 import lombok.EqualsAndHashCode;
 import lombok.ToString;
 import org.cactoos.io.InputOf;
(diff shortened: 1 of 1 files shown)


yegor256:
@rultor merge

Masynchin:
@yegor256 sorry for conflicting PR. I changed exception type in order to use `InputOf(...).stream()` and now it doesn't show any errors. `mvn clean install -Pqulice` raises error but for `RsFailure`, which I doesn't change. CI successfully builds my branch.

yegor256:
@Masynchin thanks!

Masynchin:
@yegor256 I search through Takes and Cactoos source code and found [`InputStreamOf`](https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/io/InputStreamOf.java). This is more low-level class than `InputOf`? It seems that `new InputStreamOf(...)` more suitable here than `new InputOf(...).stream()`.

I can open new PR that replaces `InputOf` to `InputStreamOf`. I also want to replace all `Files.newInputStream(file.toPath())` to `new InputStreamOf(file)`. I can't do it with `new InputOf(file).stream()` because it raises more generic `Exception` instead of `IOException`, but can with `InputStreamOf`. I am not sure about that since I am not an expert in Java, but it seems to work.

PR #1124, opened 2022-05-31
ci(rultor): remove sonar checks in release command

baudoliver7:
It is to fix sonar error about link not found while releasing with Rultor.

Ref: https://github.com/yegor256/takes/pull/1123#issuecomment-1141056834

PR #1125, opened 2022-05-31
ci(rultor): include gpg signature step

baudoliver7:
Fix: https://github.com/yegor256/takes/pull/1124#issuecomment-1142174574


yegor256:
@baudoliver7 thanks!

PR #1216, opened 2023-04-29
XeSla decimal formatting fixed

### LINKED ISSUE #1215: `XeSla` may fail depending on JVM default locale
`XeSla` internally uses `String.format("%.3f", double)` to format the server load average. 
The result is then provided as an `AttrDirective` of type `double` to be used on XML render.
The number format [should look like](https://www.w3.org/TR/1999/REC-xpath-19991116/#NT-Number) : `Number  ::=  Digits ('.' Digits?)? | '.' Digits`

The problem is that `String.format` will use the default locale to determine the decimal separator.
Hence in the environments where this separator is not `.` this class will generate malformed output.

On my machine (default = Russian)  `XeSlaTest.buildsHtmlResponse` fails, because the `Transformer` can't convert `sla` to `double`.

How about explicitly define the locale in `String.format`?

NurmukhametovAlexey:
Make `XeSla` explicitly use US locale in `String.format` for proper decimal separator.
Closes #1215

### DIFF
--- a/src/main/java/org/takes/rs/xe/XeSla.java
+++ b/src/main/java/org/takes/rs/xe/XeSla.java
@@ -24,6 +24,7 @@
 package org.takes.rs.xe;
 
 import java.lang.management.ManagementFactory;
+import java.util.Locale;
 import lombok.EqualsAndHashCode;
 import org.xembly.Directives;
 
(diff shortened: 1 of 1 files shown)


yegor256:
@rultor merge

Issue #1029, opened 2020-02-10
TkSecure returns HTTP 303 instead of HTTP 401


golszewski86:
In case of anonymous identity class org.takes.facets.auth.TkSecure throws RsForward exception with 303 See Other status.

According to [RFC2616, section10](https://tools.ietf.org/html/rfc2616#section-10), status 303 means that:

> The response to the request can be found under a different URI and SHOULD be retrieved using a GET method on that resource.

This is not the case with TkSecure because there we don't know whether the requested URI is correct, neither what HTTP method is allowed for that URI. All we know is that request requires authentication hence HTTP 401 Unauthorized should be returned.

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

0crat:
@golszewski86[/z](https://www.0crat.com/u/golszewski86) this project will fix the problem faster if you donate a few dollars to it; just [click here](https://www.0crat.com/contrib/C3FFK3YAY) and pay via Stripe, it's very fast, convenient and appreciated; thanks a lot!

PR #1240, opened 2023-09-01
Fix issue #1029 TkSecure returns HTTP 303 instead of HTTP 401

TauzhnianskyiArtem:
For [1029](https://github.com/yegor256/takes/issues/1029)

Change returning HTTP code for anonymous identity


yegor256:
@TauzhnianskyiArtem thanks for the contribution, but it seems that the quality of code is not high enough. Try to run `mvn clean -Pqulice install` and then fix the problems it reports.

yegor256:
@TauzhnianskyiArtem this is already fixed

PR #1294, opened 2024-05-02
fixes #1227 (https://github.com/yegor256/takes/issues/1227)

### LINKED ISSUE #1227: Invalid HTTP method: X-Takes-LocalAddress
I got this error sometimes with the current version of Takes :

```console
java.io.IOException: Invalid HTTP method: X-Takes-LocalAddress:
	at org.takes.rq.RqMethod$Base.method(RqMethod.java:124)
	at org.takes.facets.fallback.TkFallback.error(TkFallback.java:234)
	at org.takes.facets.fallback.TkFallback.fallback(TkFallback.java:138)
	at org.takes.facets.fallback.TkFallback.route(TkFallback.java:88)
	at org.takes.facets.fallback.TkFallback.lambda$new$0(TkFallback.java:66)
	at org.takes.tk.TkWrap.act(TkWrap.java:58)
	at org.takes.tk.TkWrap.act(TkWrap.java:58)
	at org.takes.tk.TkSlf4j.act(TkSlf4j.java:110)
	at org.takes.http.FtCli.lambda$start$0(FtCli.java:91)
	at org.takes.http.BkBasic.print(BkBasic.java:123)
	at org.takes.http.BkBasic.accept(BkBasic.java:99)
	at org.takes.http.BkSafe.lambda$new$0(BkSafe.java:46)
	at org.takes.http.BkWrap.accept(BkWrap.java:51)
	at org.takes.http.BkTimeable.accept(BkTimeable.java:92)
	at org.takes.http.BkParallel.lambda$new$0(BkParallel.java:81)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
```

While debugging, we got this:
image

It looks like `RqRequestLine.Base` doesn't do the right job :-(

EugenDueck:
When socket.getInputStream().available() returns 0, parsing of the request headers will be aborted. I guess the assumption was that available is 0 only in case of "end of stream", which is [not true](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/InputStream.html#available()):

> public int available()
>   Returns an estimate of the number of bytes that can be read (or skipped over) from this input stream without blocking, which may be 0, **or** 0 when end of stream is detected.

I think the appropriate thing to do is to get rid of the code that checks available() altogether, which is what this PR does. In order to reliably determine whether we have reached the end of the stream, we need to read(). No way around it.

### DIFF
--- a/src/main/java/org/takes/rq/RqLive.java
+++ b/src/main/java/org/takes/rq/RqLive.java
@@ -69,7 +69,7 @@ public final class RqLive extends RqWrap {
         final Collection<String> head = new LinkedList<>();
         final ByteArrayOutputStream baos = new ByteArrayOutputStream();
         Opt<Integer> data = new Opt.Empty<>();
-        data = RqLive.data(input, data, false);
+        data = RqLive.data(input, data);
         while (data.get() > 0) {
             eof = false;
             if (data.get() == '\r') {
@@ -82,11 +82,11 @@ public final class RqLive extends RqWrap {
                 if (header.has()) {
                     head.add(header.get());
                 }
-                data = RqLive.data(input, data, false);
+                data = RqLive.data(input, data);
                 continue;
             }
             baos.write(RqLive.legalCharacter(data, baos, head.size() + 1));
-            data = RqLive.data(input, new Opt.Empty<>(), true);
+            data = RqLive.data(input, new Opt.Empty<>());
         }
         if (eof) {
             throw new IOException("empty request");
@@ -170,18 +170,14 @@ public final class RqLive extends RqWrap {
      * Obtains new byte if hasn't.
      * @param input Stream
      * @param data Empty or current data
-     * @param available Indicates whether or not it should check first if there
-     *  are available bytes
      * @return Next or current data
      * @throws IOException if input.read() fails
      */
     private static Opt<Integer> data(final InputStream input,
-        final Opt<Integer> data, final boolean available) throws IOException {
+        final Opt<Integer> data) throws IOException {
         final Opt<Integer> ret;
         if (data.has()) {
             ret = data;
-        } else if (available && input.available() <= 0) {
-            ret = new Opt.Single<>(-1);
         } else {
             ret = new Opt.Single<>(input.read());
         }
--- a/src/test/java/org/takes/http/BkBasicTest.java
+++ b/src/test/java/org/takes/http/BkBasicTest.java
@@ -35,6 +35,7 @@ import java.net.ServerSocket;
 import java.net.Socket;
 import java.util.Collections;
 import java.util.concurrent.atomic.AtomicReference;
+import org.apache.commons.io.IOUtils;
 import org.cactoos.bytes.BytesOf;
 import org.cactoos.text.Joined;
 import org.hamcrest.MatcherAssert;
@@ -49,6 +50,7 @@ import org.takes.facets.fork.TkFork;
 import org.takes.misc.Href;
 import org.takes.rq.RqFake;
 import org.takes.rq.RqHeaders;
+import org.takes.rq.RqLengthAware;
 import org.takes.rq.RqPrint;
 import org.takes.rq.RqSocket;
 import org.takes.rs.ResponseOf;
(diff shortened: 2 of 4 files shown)


yegor256:
@EugenDueck I can't understand why github actions jobs are stuck. Can you please try to push something small into the branch, to restart them?

yegor256:
@EugenDueck it seems that your test is stuck. Did you run the build locally? It passes?

EugenDueck:
@yegor256  My bad! I hadn't run the other tests after fixing the issue.
So `BkBasicTest.handlesTwoRequestInOneConnection()` never finishes. I probably won't have the time to dig into it this weekend.
However, I noticed that the Content-Length of the first request in `BkBasicTest.handlesTwoRequestInOneConnection()` does not take into account the "\r\n" that `Joined()` adds after the body, so the Content-Length must be 13, rather than 11:
```java
                socket.getOutputStream().write(
                    new Joined(
                        BkBasicTest.CRLF,
                        BkBasicTest.POST,
                        BkBasicTest.HOST,
                        "Content-Length: 11",
                        "",
                        "Hello First",
                        BkBasicTest.POST,
                        BkBasicTest.HOST,
                        "Content-Length: 12",
                        "",
                        "Hello Second"
                    ).asString().getBytes()
                );
```
This still does not make the test finish. I will dig deeper when I find the time, hopefully next week.

EugenDueck:
@yegor256 

There was a deadlock in the test (the test waited for the server to close the socket, while the server waited for the client to close the socket). The commit I just pushed is supposed to fix that. With that, the tests all pass locally. As to the failing CI checks, I'm not sure how to read them. Looking at a couple of recent commits, most of them lead to failures and cancellations of one way or another.

However, looking into this issue lead me to discover a couple of issues that I would like to discuss, and therefore would like you to **not** merge this PR in its current form.

1. #649 introduced the use of `InputStream.available()` into `RqLive.parse()`, presumably as a means of checking whether the end of the stream has been reached (i.e. the other side has closed its side of the connection). Unfortunately, InputStream.available() can't be relied upon for that - see the quote in my first message of this PR as to why. So what can happen, and what actually does happen in #1227, hence this PR, is that RqLive.parse() stops parsing the request, because available() returns 0. So this needs to be fixed, even though up to now, we have only seen the bug manifest on Windows.

2. Also in #649, `InputStream.available()` was added to `BkBasic.accept()`, again, in order to check whether there is more data to come or not. This is after a request has been finished, waiting for the next request. So this means that Takes keeps the connection open only as long as the client can keep up "http pipelining" requests into the connection fast enough. The moment the client stops for just a tiny bit, Takes will disconnect. Which all but defeats the purpose of persistent connections. Http server implementations generally have a [keep-alive timeout](https://nginx.org/en/docs/http/ngx_http_core_module.html#keepalive_timeout) for persistent connections that defaults to anywhere between a couple of seconds and about 1 minute. Using .available(), the timeout is effectively "0". The tcpdump in the following screenshot, taken while doing

`(echo -ne "GET /a HTTP/1.1\r\nHost: localhost:8100\r\n\r\n" ; cat) | nc localhost 8100`

shows the immediate closure of the connection by Takes:

image

3. While looking into these issues, and wondering why BkBasicTest.handlesTwoRequestInOneConnection() succeeded even though the content length was wrong, I discovered #1295

I don't feel like it is a good idea to fix 1., without also addressing 2. I would like to have a reliable "persistent connection" implementation first, with tests that show that it works, before fixing this bug, while making sure that persistent connections still work.

Actually, my preferred approach would be to proceed in this order:
1. remove support for persistent connections altogether (effectively rolling back #649) 
2. be happy (it will automatically fix this bug, but I would want to add a regression test to make sure it does and stays that way)
3. create a fresh "persistent connections" feature request for anyone willing to take up

Why do I suggest removing the persistent connections feature? Because that feature is barely working, as the tcpdump shows. And fixing the current implementation is not a trivial matter, in particular due to the need to implement a keep-alive timeout to prevent indefinite blocking of the connection and thread, which means coordinating with at least one more thread - unless of course we go NIO, which however would amount to a complete rewrite of the whole framework.

So we would just remove something that cannot be used in its current form, cannot be fixed easily, but introduces problems.

What do you think?

Issue #1301, opened 2024-06-21
RqLengthAware wrongly assumes available gives a good indication about number of bytes to read when no content length is aviable


laeubi:
`RqLengthAware` uses `available` in case of no content length is found, this is simply wrong, according to the docs this:

> Returns an estimate of the number of bytes that can be read (or skipped over) from this input stream without blocking, which may be 0, or 0 when end of stream is detected.

The default implementation even return always 0....

instead it should simply return the stream as is if the content length can't be computed!

Issue #1116, opened 2022-01-14
org.takes.Request.body().read() never returns -1 to indicate end of request body


golszewski86:
Consider the following example of Takes server and HTTP made to it:

```java
package takes.test;

import com.jcabi.http.request.JdkRequest;
import java.io.IOException;
import java.io.InputStream;
import java.net.SocketTimeoutException;
import java.time.Duration;
import java.util.concurrent.atomic.AtomicBoolean;
import lombok.AllArgsConstructor;
import lombok.extern.slf4j.Slf4j;
import org.takes.Request;
import org.takes.Response;
import org.takes.Take;
import org.takes.http.Exit;
import org.takes.http.FtBasic;
import org.takes.rs.RsEmpty;

/**
 *
 * @since 1.0
 */
@Slf4j
public class TakesTimeoutTest {

    public static void main(final String[] args) throws Exception {
        Thread.currentThread().setName("Client");
        final AtomicBoolean done = new AtomicBoolean(false);
        final Thread server = new Thread(new Server(done::get), "Server");
        server.start();
        final int timeout = (int) Duration.ofSeconds(5).toMillis();
        try {
            new JdkRequest("http://localhost:8080/")
                .method("POST")
                .body().set("bar").back()
                .timeout(timeout, timeout)
                .fetch();
            log.info("Received response");
        } catch (final IOException exc) {
            if (exc.getCause() instanceof SocketTimeoutException) {
                log.error("Request timed out!");
            } else {
                throw exc;
            }
        } finally {
            done.set(true);
        }
        server.join();
    }

    @AllArgsConstructor
    private static final class Server implements Runnable {

        private final Exit exit;

        @Override
        public void run() {
            try {
                log.info("Starting server on port 8080");
                new FtBasic(
                    new Take() {
                        @Override
                        public Response act(final Request req) throws Exception {
                            log.info("Received request");
                            final InputStream body = req.body();
                            int character;
                            while ((character = body.read()) >= 0) {
                                log.info("Read: {}", (char) character);
                            }
                            log.info("Returning response");
                            return new RsEmpty();
                        }
                    }, 8080
                ).start(this.exit);
            } catch (final Exception exc) {
                log.error("Server exception", exc);
            } finally {
                log.info("Server stopped");
            }
        }
    }
}
```
### What it does

* It starts a Takes HTTP server on port 8080
* The server accepts any HTTP requests, logs information about received request, logs every character from the request body (assuming it is a plain text body), and logs information about returned response
[...]

PR #1302, opened 2024-06-21
Only cap the input if content length header is given

laeubi:
Currently RqLengthAware uses the InputStream#available to cap the input if no Content-Length header is given, but the return value does not say anything about the real content length, only how many bytes can be read without blocking (what might be 0).

This now only returns a CapInputStream when a Content-Length is given in the request.

Fixes https://github.com/yegor256/takes/issues/1301

### DIFF
--- a/src/main/java/org/takes/rq/RqLengthAware.java
+++ b/src/main/java/org/takes/rq/RqLengthAware.java
@@ -51,6 +51,10 @@ import org.takes.Request;
  */
 @EqualsAndHashCode(callSuper = true)
 public final class RqLengthAware extends RqWrap {
+    /**
+     * Constant for the Content-Length header.
+     */
+    private static final String CONTENT_LENGTH = "Content-Length";
 
     /**
      * Ctor.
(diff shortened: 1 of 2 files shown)


laeubi:
@yegor256 anything else needed here?

yegor256:
@laeubi would be great if you can reproduce the problem with a unit test too. Possible?

laeubi:
I now added a testcase that fails without my patch but succeeds when the patch is applied.

yegor256:
@laeubi now looks perfect, but I'm afraid there is a formatting mistake in the unit test (see what Qulice is saying).

laeubi:
Is there a way to apply the desired formatting automatically?

yegor256:
@laeubi we don't have that, here is why: https://www.yegor256.com/2018/01/16/educational-aspect-of-static-analysis.html

laeubi:
> @laeubi we don't have that, here is why: https://www.yegor256.com/2018/01/16/educational-aspect-of-static-analysis.html

While theoretically this might be good I contribute to literally hundreds of projects each using slightly different "better styles" so learning all of them is at least not really affordable for contributors especially now I have to count whitespace, rerun build again and even decode non trivial instructions like 

> Indentation (3) must be same or less than previous line (2), or bigger by exactly 4 (CascadeIndentationCheck)

that's really a mess and don't bring any value (neither to me nor to the project) as it is simply stupid work that can be automated.

yegor256:
@laeubi I agree. However, we don't have a tool for that. We tried to create it in the past, but failed :( Sorry.

laeubi:
I have now addressed the style errors, I know at least Apache uses https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md

yegor256:
@laeubi it seems that something is wrong with the unit test introduced, since all workflow runs are hanging. Can you please check?

laeubi:
> At laest in the UI they all run fine and fast...

grafik

Also in github I see they pass:

> [INFO] Running org.takes.rq.RqLengthAwareTest
> [INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s -- in org.takes.rq.RqLengthAwareTest

what seems hanging is the proxy test

> [INFO] Running org.takes.tk.TkProxyTest
> 07:27:54.309 [main] INFO com.jcabi.http.request.BaseRequest -- #fetch(POST localhost:40177 /а): [200 OK] in 1ms

it also hangs locally for me I suspect this is due to this bug:
- https://github.com/yegor256/takes/issues/1116

before my change there was **always** a CapInputStream returned that terminated (but maybe with wrong bytes) always, now it returns the raw input stream that possibly never returns...

laeubi:
Yes I can confirm it hangs in socket read:
grafik

PR #1314, opened 2024-08-02
Only print body if there is a body

laeubi:
Currently reading from a request can hang forever if there is no content available (e.g. GET / HEAD / DELETE / ...).

This adds a check if any of the content headers are there to indicate that actually something can be read.

Follow-Up for
- https://github.com/yegor256/takes/pull/1302


laeubi:
Okay that don't work, because it seems some tests write a body without any content headers (what theoretically is allowed for HTTP 1.0 on connection close)... I'll have to think a little bit more about this.

yegor256:
No CI has ever run on this PR (it's from 2024 and predates the current workflows), so I can't confirm it's still green or applies cleanly. Flagging for a manual look/re-run before merging.

---

PR #1335, opened 2024-10-04
Changed CapInputStream

UFA-MOROZOV:
@yegor256, after reviewing code of CapInputStream i found possible mistake in logic. There we have skip function, which skips buffer, however this class is different from InputStream because of its cap. So by logic, skipping something should also change cap. So i ran all tests written before me with my changes of code, they all went good, and i created my own test, that checks one simple case:
we have stream and cap of 50
we skip 75, it should skip 50, since the cap is 50, and also it should decrease "more" variable, which represents how much we can read at that point and not violate cap.


yegor256:
@rultor merge

yegor256:
@UFA-MOROZOV thanks!

PR #1346, opened 2024-10-29
Do not allow closing of the live input stream

laeubi:
Currently the raw stream is passed to downstream consumers of RqLive, this has the problem that such consumers can close the stream. In case of a socket, this means the socket itself is closed and no response can be send by takes.

This wraps the original live input stream into one that simply ignores the close request.

### DIFF
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -37,6 +37,7 @@ import org.takes.Request;
 import org.takes.Response;
 import org.takes.Take;
 import org.takes.rq.RqLive;
+import org.takes.rq.RqUncloseable;
 import org.takes.rq.RqWithHeaders;
 import org.takes.rs.RsPrint;
 import org.takes.rs.RsText;
@@ -120,7 +121,7 @@ public final class BkBasic implements Back {
     private void print(final Request req, final OutputStream output)
         throws IOException {
         try {
-            new RsPrint(this.take.act(req)).print(output);
+            new RsPrint(this.take.act(new RqUncloseable(req))).print(output);
         } catch (final HttpException ex) {
             new RsPrint(BkBasic.failure(ex, ex.code())).print(output);
             // @checkstyle IllegalCatchCheck (10 lines)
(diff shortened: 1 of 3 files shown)


laeubi:
Another option would be to handle it on all different places where such streams are used, but it seems more clean to handle it more like at the roots.

laeubi:
One problem is that currently `RqTemp` uses `RqLive` in a way that relies on close... so one can also handle this in `BkBasic` 

@yegor256 any guidance on this how it is best to be handled? Maybe have (another) wrapper `RqNoClose` ?!?

laeubi:
Testcase is currently also missing. Where would be the best place to place such test that fails if no response is given?

yegor256:
@laeubi who is the consumer of the InputStream that may close it "accidentally"? Can you give an example?

laeubi:
In my case I have a POST handler with a `TkRegex`, there I read the body of the message with a function that closes the passed in stream. While I could workaround this there, it seems consumers should not be able to close the stream/socket as otherwise sending a response is no longer possible.

yegor256:
@laeubi good point, I believe, you're right. However, I would implement it slightly differently. In the `BkBasic` I would replace this line:

```
new RsPrint(this.take.act(req)).print(output);
```

With this one:

```
new RsPrint(this.take.act(new RqUncloseable(req))).print(output);
```

It seems that this is what you suggested earlier as an alternative.

laeubi:
@yegor256 it seems the copyright check fails because it explicitly wants something like 

> Punch line is: "Copyright (c) 2014-2024 Yegor Bugayenko"

is this intentional? Should other names be allowed? I'm not that familiar with MIT license in this regard.

yegor256:
Please fix the merge conflicts so this pull request can be merged.

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

PR #1347, opened 2024-10-29
Make sure all bytes written are flushed

laeubi:
Currently BkBasic uses a BufferedOutputStream, so all content is first buffered before send to the actual stream. While the stream is implicitly flushed after writing headers, it is not when writing the body. So if the body is quite small it might never been send to the socket before close.

This now explicitly flush the content after calling print to not loose bytes regardless of how the printing is actually implemented.


laeubi:
@yegor256 can you review this?

PR #1359, opened 2024-11-25
Dont set init capacity of HashMap to 0

daniilzimin4:


yegor256:
@daniilzimin4 what is the motivation for this?

PR #1376, opened 2024-12-28
adding uri handler for hiding null checks from code

Kartofanych:


Kartofanych:
@yegor256 Hey! Can u check please my idea to hide nulls inside helper class that works with URI? Also i think it should have public static methods, because it is just helper class and it doesnt have any variables and logic.

yegor256:
@Kartofanych the idea is interesting indeed, but may only work if we implement this URI handler in a separate library. Here, it only tries to hide the fact: NULL is returned by JDK class. No need to hide it.

PR #1418, opened 2025-07-14
Removed depency with critical vulnerability

CacucoH:
Recently, in `pom.xml` I noticed critically vulnerable dependency that may expose entire app. I also noticed that modern logging service _(SLF4J)_ so the vulnerable Log4j even is not required here

Therefore, I've deleted those files to improve security


yegor256:
@CacucoH Thanks for your contribution, we appreciate it! Before we can merge this pull request, we need to see all CI workflows pass without errors. We can't merge the pull request into the master branch unless all workflows are green. Try to fix them, also paying attention to code style issues.

CacucoH:
@yegor256  I've searched through the code and noticed that log4j is not used. You are already use a modern SLF4j logger instead. 

Log4j is deprecated and is no longer used, since it contains severe vulnerabilities (there is vulnerability rated as 9.8 for CVSS score). It may lead to users' personal data leakage, or even worse, an RCE

CacucoH:
@yegor256  seems like now all tests are passed and `takes`  works fine. I don't see any problems with removing an EXTREME vulnerable module from your code, especially if there is another logging lib that is already used (SLF4J)

yegor256:
@rultor merge

yegor256:
@CacucoH thanks!

Issue #1441, opened 2025-07-25
Unhandled scope of http exceptions


XriXis:
On a control char in the request, instead of response with 400 code thread with server just lies down, because HttpException thrown at BkBasic#accept, while initializing arguments for the BkBasic#print, where HttpException s are actually handled

Error with 400 response code occurs on RqLive construction. While general handling of http errors, and creation of corresponding response occurs in the method, that called after (and require) RqLive initialisation

Problematic lines are marked by red background:

<img width="580" height="900" alt="Image" src="https://github.com/user-attachments/assets/1fd6108e-8f8e-42d1-8f3b-27c7ab35ae21" />

Stack trace on receiving "GET /\n":
<code>Exception in thread "Thread-0" java.lang.IllegalStateException: org.takes.HttpException: [400] illegal character 0x0A in HTTP header line #1: "GET /"
	at org.takes.http.BkBasicTest.lambda$returnsABadRequestToAnInvalidRequestUri$6(BkBasicTest.java:358)
	at java.base/java.lang.Thread.run(Thread.java:1570)
Caused by: org.takes.HttpException: [400] illegal character 0x0A in HTTP header line #1: "GET /"
	at org.takes.rq.RqLive.legalCharacter(RqLive.java:137)
	at org.takes.rq.RqLive.parse(RqLive.java:69)
	at org.takes.rq.RqLive.<init>(RqLive.java:37)
	at org.takes.http.BkBasic.accept(BkBasic.java:80)</code>

yegor256:
@XriXis it's fixed already, right?

Issue #1058, opened 2020-07-25
"get \" produces StackOverflowError


rusefillc:
Below shell snippet hopefully tells the complete story

```telnet localhost 8001
Trying 127.0.0.1...
Connected to localhost.localdomain.
Escape character is '^]'.
get \

HTTP/1.1 500 Internal Server Error
Content-Length: 50287
Content-Type: text/plain

java.lang.StackOverflowError
        at java.base/java.lang.Exception.<init>(Exception.java:66)
        at java.base/java.net.URISyntaxException.<init>(URISyntaxException.java:62)
        at java.base/java.net.URI$Parser.fail(URI.java:2913)
        at java.base/java.net.URI$Parser.parseAuthority(URI.java:3247)
        at java.base/java.net.URI$Parser.parseHierarchical(URI.java:3158)
        at java.base/java.net.URI$Parser.parse(URI.java:3114)
        at java.base/java.net.URI.<init>(URI.java:600)
        at org.takes.misc.Href.createUri(Href.java:301)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
        at org.takes.misc.Href.createUri(Href.java:313)
[...]

PR #1442, opened 2025-07-25
#1058: BkBasic returns 400 Bad Request for malformed URIs

XriXis:
This PR solves issue #1058 and completes the corresponding @todo in `BkBasicTest`.

### Changes introduced:
1. Rewrote `Href#createUri` to reject malformed URIs containing illegual characters before path, which previously led to incorrect parsing and silent URI normalization.
2. Updated `BkBasic#print()` to handle `IllegalArgumentException` and return HTTP 400 (Bad Request).
3. Re-enabled, validated, and renamed the test `returnsABadRequestToAnInvalidRequestUri` (renamed to `returnsABadRequestToAMissingPath`), which now passes.
4. Add new test, assosiated with vulnerable for stability bug #1441
All tests pass locally.

### DIFF
--- a/src/main/java/org/takes/http/BkBasic.java
+++ b/src/main/java/org/takes/http/BkBasic.java
@@ -100,18 +100,24 @@ public final class BkBasic implements Back {
     @SuppressWarnings("PMD.AvoidCatchingThrowable")
     private void print(final Request req, final OutputStream output)
         throws IOException {
+        new RsPrint(handledResponse(req)).print(output);
+    }
+
+    private Response handledResponse(Request req){
         try {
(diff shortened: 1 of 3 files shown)


yegor256:
@rultor merge

yegor256:
@XriXis many thanks for your contribution! Next time, please, ping me. Otherwise, I simply miss a PR.

XriXis:
@yegor256 many thanks for accepting my changes. And yes, I have read this article and I have wrote you PM in telegram (same alias, as here - xrixis) to notify you about complited work, but message seems was accidentionally lost. Anyway, I will appreciate a lot, if that message will be read

Issue #1435, opened 2025-07-24
Refactor TkHtmlTest to improve structure and edge case coverage (relates to #1408)


niotu:
Improve and Refactor `TkHtmlTest` for Maintainability and Coverage (related to [#1408](https://github.com/yegor256/takes/issues/1408))

The current implementation of [`TkHtmlTest`](https://github.com/yegor256/takes/blob/master/src/test/java/org/takes/tk/TkHtmlTest.java) can be improved in the following ways:

1. **Duplication**: Each test manually builds a nearly identical expected response using `Joined(..., body)`. This can be extracted into a helper method or reused variable.

2. **Edge Cases Missing**:
   - No tests for null or empty input.
   - No assertion that `Content-Type` and `Content-Length` headers are semantically correct (not just matched by string).

3. **Structure Validation**:
   - Could add a test to verify that the body actually starts/ends with `<html>` tags to ensure semantic HTML is passed.

4. **Test Readability**:
   - Introduce `@DisplayName` annotations or comment grouping to make intent clearer.

A can provide a PR if the maintainers agree.

yegor256:
Yes, please go ahead with the pull request. Keep one assertion per test and phrase failure messages negatively, as the rest of the suite does.

PR #1487, opened 2025-10-30
enhance TkHtml test coverage and readability

eshabakhov:
Pull request solves https://github.com/yegor256/takes/issues/1435

### DIFF
--- a/src/test/java/org/takes/tk/TkHtmlTest.java
+++ b/src/test/java/org/takes/tk/TkHtmlTest.java
@@ -4,7 +4,6 @@
  */
 package org.takes.tk;
 
-import java.io.InputStream;
 import org.cactoos.io.InputStreamOf;
 import org.cactoos.iterable.IterableOf;
 import org.cactoos.text.Joined;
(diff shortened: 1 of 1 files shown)


src/test/java/org/takes/tk/TkHtmlEmptyTest.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Test case for {@link TkHtml with empty body}.
+ * @since 0.10
+ */
+final class TkHtmlEmptyTest {
+
+    @Test
+    @DisplayName("Throw IllegalStateException when string body is null")
+    void failsOnNullInputString() {
+        final String body = null;
```
sopronyr: Why you add final to String? String is final by default. It look s like mess

eshabakhov: @sopronyr What do you mean when you say that string is final by default? Final for a string variable means that the String object is immutable, and the reference to this object **cannot be redefined after initialization**. 

The static analyzer also does not skip classes, fields and variables without final, btw..

eshabakhov:
@yegor256 thanks for the article!
I thought it would be preferable to use fewer methods in tests so that qulice wouldn't complain. 

Fixed, check please. Is it correct to check for null in tests like this?

Issue #1440, opened 2025-07-25
Spaces in the request divide url instead of proper substitution causing


XriXis:
On attempt to cause char substitution (' ' to %20 or '\\' to %5C) in a mock test I observed partial lost of char sequence:

socket.getOutputStream().write(
  new RqPrint(
    new RqFake("GET", "/?u=Hello World")
  ).asString().getBytes()
);

On the RqHref#href i cought following context (uri variable just eliminate part of the request):

<img width="881" height="496" alt="Image" src="https://github.com/user-attachments/assets/4d9d0b57-1e35-454b-8059-bade341a651c" />

PR #1536, opened 2025-12-24
#1440 Test to reproduce bug behavior: Space truncation in rq

vladigeras:
Capture the current behavior in tests. To reproduce https://github.com/yegor256/takes/issues/1440

### DIFF
--- a/src/test/java/org/takes/rq/RqHrefTest.java
+++ b/src/test/java/org/takes/rq/RqHrefTest.java
@@ -194,8 +194,11 @@ import org.takes.HttpException;
 
     /**
      * Space truncation.
-     * todo: #1440 There is a space truncation in request. The string is being truncated after
-     * the space because the URI is not properly encoded.
+     * todo: #1440 This test doesn't work because there is a space truncation in request.
+     * The string is being truncated after the space because the URI is not properly encoded.
+     * Let's make it work by doing URI encoding, because URL encoding converts characters into
+     * a format that can be transmitted over the Internet. URL encoding replaces unsafe ASCII
+     * characters with a "%" followed by two hexadecimal digits.
      */
     @Disabled
     @Test
--- a/src/test/java/org/takes/rq/RqRequestLineTest.java
+++ b/src/test/java/org/takes/rq/RqRequestLineTest.java
@@ -164,8 +164,11 @@ import org.takes.HttpException;
 
     /**
      * Space truncation.
-     * todo: #1440 There is a space truncation in request. The string is being truncated after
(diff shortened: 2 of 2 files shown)


yegor256:
@vladigeras this is not exactly how unit testing works:

> Later, when the time comes to fix the bug, these tests will fail, signaling that they need to be updated. 

When a test fails it's a signal that the code is broken. Not the test.

vladigeras:
@yegor256 these are so-called characterization tests (also known as Golden Master Testing). They capture the current (flawed) behavior to establish a baseline. When the bug is eventually fixed, these tests should fail, confirming the behavior changed. This doesn’t violate unit testing principles; it’s a controlled way to document known issues until they’re addressed.

Although this approach is typically used in legacy projects (and **I don't consider this service to be one**), I still find the practice useful.

yegor256:
@vladigeras it's an interesting approach, but we practice unit testing differently. I suggest implementing the tests such that they fail. Then, annotate them with `@Disable` and attached puzzles, see: https://www.yegor256.com/2010/03/04/pdd.html

vladigeras:
@yegor256 This is an interesting approach too. I rewrote according to it.

> Nevertheless, your goal of working with your specific task is to finish it as soon as possible and reintegrate your branch into master.

In fact, it's exactly what I'm doing. Relative to PDD, it appears to be the opposite. PDD includes an aspect: leaving behind a test that compiles but never actually executes. This is somewhat similar to keeping commented-out code or unreachable code in the codebase. 
My approach, on the other hand, includes another aspect: it has retains a working and executable test that already captures the system's current behavior. Even if, under good code coverage, it might be somewhat redundant.

vladigeras:
@yegor256 
1) One of my iteration contains `@todo`, but pipeline failed in pdd task https://github.com/yegor256/takes/actions/runs/20499906967/job/58905628988 with comment:
```
ERROR: ERROR: src/test/java/org/takes/rq/RqHrefTest.java; PDD::Error at src/test/java/org/takes/rq/RqHrefTest.java:197: @todo found, but puzzle can't be parsed, most probably because @todo is not followed by a puzzle marker, as this page explains: https://github.com/cqfn/pdd#how-to-format
If you can't understand the cause of this issue or you don't know how to fix it, please submit a GitHub issue, we will try to help you: https://github.com/cqfn/pdd/issues. This tool is still in its beta version and we will appreciate your feedback. Here is where you can find more documentation: https://github.com/cqfn/pdd/blob/master/README.md.
```
so I change it to `todo`. Is it still wrong? 

2) Done

PR #1569, opened 2026-04-10
Polish readme grammar

artur-intech:
P.S. Let it be my low-hanging fruit.


yegor256:
@artur-intech I don't see a difference, but thanks!)

artur-intech:
@yegor256

The difference is in grammar: The "whys" are not real questions, but statements. Inversion does not apply in this case. No question mark is needed either.

The rest has just been rephrased for a more natural flow.

PR #1634, opened 2026-05-11
Fix constructor parameter order in TkMethods to match FkMethods (#1005)

### LINKED ISSUE #1005: Inconvenient TkMethods and FkMethods constructor parameters order
There're two classes `TkMethods` and `FkMethods`. First has constructor:
```java
TkMethods(final Take take, final String... methods)
```
second:
```java
FkMethods(final Collection<String> mtds, final Take tke)
```
Both accepts `Take` and list of methods to support, but in `TkMethods` takes parameter is first and in `FkMethods` it's second. I always mistype these arguments when using takes and try to pass take as second parameter to `TkMethods`. I'd change `TkMethods` (since it's a wraper based on `FkMethods`, and `FkMethods` is primary class here) to:
```java
TkMethods(final Collection<String> methods, final Take take)
```
same as in `FkMethods`.

kreinba:
Closes #1005.

The root cause is that `TkMethods(Take, String...)` puts `Take` first while `FkMethods(Collection<String>, Take)` puts the method collection first. Anyone who learns the API from `FkMethods` will instinctively swap the arguments when calling `TkMethods`, getting a confusing compile error. This PR adds a new primary constructor `TkMethods(Collection<String>, Take)` that matches the `FkMethods` signature, and rewires the existing `TkMethods(Take, String...)` constructor to delegate to it so that all existing callers continue to compile unchanged.

https://claude.ai/code/session_01C4kW3Zay6pr3EhPAJ2kDQx

### DIFF
--- a/src/main/java/org/takes/facets/fork/TkMethods.java
+++ b/src/main/java/org/takes/facets/fork/TkMethods.java
@@ -5,6 +5,7 @@
 package org.takes.facets.fork;
 
 import java.net.HttpURLConnection;
+import java.util.Collection;
 import lombok.ToString;
 import org.cactoos.list.ListOf;
 import org.takes.HttpException;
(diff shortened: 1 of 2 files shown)


kreinba:
@g4s8 All CI checks are green. The fix adds a second constructor `TkMethods(Collection<String>, Take)` whose parameter order matches `FkMethods(Collection<String>, Take)`, making both constructors consistent. The existing `TkMethods(Take, String...)` constructor now delegates to the new one. A new test `acceptsCollectionOfMethodsBeforeTakeMatchingFkMethods` exercises the new constructor directly. Please review when you have a moment.

PR #1684, opened 2026-05-26
#1440: Encode query chars in request target

### LINKED ISSUE #1440: Spaces in the request divide url instead of proper substitution causing

GHX5T-SOL:
/claim #1440

Fixes #1440.

Changes:
- Percent-encode raw spaces and backslashes inside query values before the request target is used by `RqHref`.
- Keep malformed request lines with extra elements, extra version spacing, and invalid path-only targets rejected.
- Serialize `Href` spaces as `%20` instead of `+`.

Validation:
- `git diff --check` -> passed.
- `mvn -q -Dtest=RqMethodTest,BkBasicTest,RqRequestLineTest,RqHrefTest,HrefTest test` -> passed.
- `mvn --errors --batch-mode -Pqulice qulice:check` -> BUILD SUCCESS.
- `mvn --errors --batch-mode clean install -Pqulice -Pdeep` -> unit tests passed: 666 run, 0 failures, 0 errors, 5 skipped; invoker `file-manager` passed; failed later in local `revapi` plugin with missing `org/apache/logging/log4j/spi/LoggerAdapter`.
- Current GitHub checks fail during action download from `codeload.github.com` before project commands run; rerun requires repository admin rights.

### DIFF
--- a/src/main/java/org/takes/misc/Href.java
+++ b/src/main/java/org/takes/misc/Href.java
@@ -309,7 +309,7 @@ public final class Href implements CharSequence {
         try {
             return URLEncoder.encode(
                 txt, Charset.defaultCharset().name()
-            );
+            ).replace("+", "%20");
         } catch (final UnsupportedEncodingException ex) {
             throw new IllegalStateException(
                 String.format("Failed to encode '%s'", txt),
--- a/src/main/java/org/takes/rq/RqRequestLine.java
+++ b/src/main/java/org/takes/rq/RqRequestLine.java
@@ -6,6 +6,9 @@ package org.takes.rq;
 
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.EqualsAndHashCode;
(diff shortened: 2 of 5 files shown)


akmhatey-ai (approved):
Checked current head b0dfb7a1326af3d854ab2f6ebd7cc777bf665f3b against #1440.

The request-target parsing now preserves query text containing spaces by encoding illegal query characters instead of truncating after the first raw space. I also checked the ambiguous case where a query value itself contains text like HTTP/1.1: the parser treats the final HTTP/x.y token as the version and percent-encodes the earlier space inside the query target, which is lenient but matches this fix's goal and is safer than dropping the tail.

Validation:
- git diff --check origin/master...HEAD passed.
- Diff-only Gitleaks over origin/master...HEAD found no leaks.
- Hosted Maven/check matrix is green, including Ubuntu, Windows, and macOS Java 23.

Limitations: this Windows host has no local java, javac, mvn, or Maven wrapper available, so I did not rerun Maven locally. One compatibility note is that Href.encode() now emits %20 rather than + for spaces, which matches URI percent-encoding but differs from form-encoding conventions. I do not see that as a blocker for this issue.

PR #1686, opened 2026-05-26
#1114: configure duplicate finder in qulice

### LINKED ISSUE #1114: pom.xml:392-395: duplicate-finder-maven-plugin found...
The puzzle `958-e512fbf3` from #958 has to be resolved: 

https://github.com/yegor256/takes/blob/72f09dd348a86f566fd627681e7d30489f83c148/pom.xml#L392-L395

The puzzle was created by @baudoliver7 on 30-Nov-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

GHX5T-SOL:
Fixes #1114.

Summary:
- remove the standalone duplicate-finder plugin and its stale PDD block
- keep the same ignored dependencies through Qulice duplicatefinder excludes

Validation:
- `mvn --errors --batch-mode -Dstyle.color=never -DskipTests -Pqulice qulice:check` -> `BUILD SUCCESS`
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk-amd64 mvn --errors --batch-mode -Dstyle.color=never -DskipTests -Drevapi.skip=true clean install -Pqulice -Pdeep` -> `BUILD SUCCESS`
- `git diff --check` -> passed
- `git diff --no-ext-diff | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks

Limitation:
- The same `clean install -Pqulice -Pdeep` command without `-Drevapi.skip=true` compiled, built both jars, and passed the invoker build, then failed in Revapi with external-class-exposed-in-API reports plus a local Maven wagon `LoggerAdapter` classloading error. I left that as a local validation limitation because this patch only removes the standalone duplicate-finder configuration and the Qulice duplicatefinder path passes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -656,31 +656,6 @@
           </annotationProcessorPaths>
         </configuration>
       </plugin>
-      <plugin>
-        <groupId>org.basepom.maven</groupId>
-        <artifactId>duplicate-finder-maven-plugin</artifactId>
-        <version>2.0.1</version>
-        <configuration>
-          <ignoredDependencies>
-            <dependency>
-              <groupId>javax.json</groupId>
-              <artifactId>javax.json-api</artifactId>
-              <version>1.1.4</version>
-            </dependency>
-            <dependency>
-              <groupId>javax.xml.bind</groupId>
-              <artifactId>jaxb-api</artifactId>
-              <version>2.3.1</version>
-            </dependency>
-          </ignoredDependencies>
-        </configuration>
-        <!--
-        @todo #958:30min duplicate-finder-maven-plugin found module-info as duplicate
-         and different class in [javax.json:javax.json-api:1.1,javax.xml.bind:jaxb-api:2.3.0].
-         After resolving duplicatefinder in Qulice, remove this plugin and configure qulice
-         plugin to enable these exclusions (see at https://www.qulice.com/qulice-maven-plugin/example-exclude.html).
-        -->
-      </plugin>
     </plugins>
   </build>
   <profiles>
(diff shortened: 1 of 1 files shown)


akmhatey-ai (approved):
LGTM, approving at head ff7558b7595ba90507218c801e1e818662fd3ecf.

This resolves #1114 / puzzle 958-e512fbf3 by removing the stale standalone duplicate-finder-maven-plugin block and moving the two known duplicatefinder exceptions into Qulice. The important part is that this narrows the old broad duplicatefinder:.* exclusion to the two specific dependencies, so duplicatefinder coverage is restored for the rest of the dependency graph.

I checked that the PDD text and standalone plugin are gone from pom.xml, pom.xml still parses, the Qulice duplicatefinder excludes are exactly javax.json:javax.json-api:1.1.4 and javax.xml.bind:jaxb-api:2.3.1, git diff --check passes, and diff-only Gitleaks found no leaks. I also found no same-scope competing open PR in the duplicate-finder/#1114 search. I could not rerun Maven locally on this Windows host because Java/Maven are unavailable, so I am relying on the successful hosted Maven/Qulice checks across Ubuntu, Windows, and macOS for build-level validation.

yegor256:
Merging this pull request as all CI checks pass, it has an approved review, and the change is within the auto-merge size limit.

PR #1733, opened 2026-08-12
fix: Put HTTP print methods behind a contract

### LINKED ISSUE #1007: Every public method in a good object should implement his counterpart from an interface
According to Yegor's article https://www.yegor256.com/2014/11/20/seven-virtues-of-good-object.html#2-he-works-by-contracts :
The rule here is simple: every public method in a good object should implement his counterpart from an interface. If your object has public methods that are not inherited from any interface, he is badly designed.

Looking through the code I have found at least the following classes which break this rule:
- [RqPrint](https://github.com/yegor256/takes/blob/master/src/main/java/org/takes/rq/RqPrint.java)
- [RsPrint](https://github.com/yegor256/takes/blob/master/src/main/java/org/takes/rs/RsPrint.java)

mvanhorn:
Add one shared `org.takes.Printable` production interface that declares the six existing printing operations with their current return types, parameters, and checked-I/O behavior. Make `RqPrint` and `RsPrint` implement that interface alongside `Text`, marking each existing printing operation as an override without changing serialization, validation, buffering, or stream ownership behavior. Use the shared contract rather than parallel request/response interfaces because the two classes already expose an identical operation set, while `Text` remains the composable representation used by current callers. `RqPrint` and `RsPrint` expose the same six public printing operations for a complete HTTP message, its head, and its body, either as a string or into an `OutputStream`. Although both decorators now implement Cactoos `Text`, only `asString()` belongs to that contract; the public `print`, `printHead`, and `printBody` overloads remain class-specific methods. This conflicts with the repository's documented architecture that components expose behavior through composable interfaces. The issue is limited to the two concrete decorators it identifies and does not call for a repository-wide object audit.

Testing: A request printer referenced as `Printable` renders the complete request, head, and body with the same CRLF boundaries and content as `RqPrint` does today; A response printer referenced as `Printable` renders the complete response, head, and body with the same status/header validation and UTF-8 output as `RsPrint` does today.

Fixes #1007

### DIFF
--- a/src/main/java/org/takes/Printable.java
+++ b/src/main/java/org/takes/Printable.java
@@ -1,0 +1,61 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2014-2026 Yegor Bugayenko
+ * SPDX-License-Identifier: MIT
+ */
+package org.takes;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * Printable HTTP message.
+ *
+ * <p>An object implementing this interface can print a complete HTTP message,
+ * its head, or its body into a string or an output stream.
+ *
+ * @since 2.0
+ */
+public interface Printable {
+
+    /**
+     * Print the complete HTTP message into a string.
+     * @return Complete HTTP message
+     * @throws IOException If something goes wrong
+     */
+    String print() throws IOException;
+
+    /**
+     * Print the complete HTTP message into an output stream.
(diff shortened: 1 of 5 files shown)


yegor256:
Holding off — CI is not fully green, and this adds a new public `Printable` contract across `RqPrint`/`RsPrint`, which is an API-shape change worth a maintainer's eyes rather than an automated merge.

---

